PR #161, opened 2024-06-26
Add `Deployer.prev` function

duncancmt:


dekz (approved):
LGTM, maybe throw in a unit test for `prev` (also `next`) functionality

PR #177, opened 2024-07-17
Linea, Again

duncancmt:


sh/address.sh
```diff
@@ -172,7 +172,11 @@ function create3 {
     declare -r salt
 
     # this is the "create3" shim inithash
-    declare -r inithash='0x3bf3f97f0be1e2c00023033eefeb4fc062ac552ff36778b17060d90b6764902f'
+    if [[ $(get_config isShanghai) != [Tt]rue ]] ; then
```
iolloyd, line 175: perhaps use the posix way of converting `tr '[:upper:]' '[:lower:]'`

duncancmt: thank you so much for the review!

I've gotten into the habit of comparing against `[Tt]rue` to deal with JSON/JS and Python differences. but actually, I think this would be fine with just `true`? mostly I'm trying to avoid the footgun of accepting a badly-encoded config file

PR #231, opened 2024-09-08
UniV4

duncancmt:


src/core/UniswapV4.sol
```diff
@@ -0,0 +1,770 @@
...
+    /// bytes that are not consumed in the decoding process. The first byte of `data` describes
+    /// which of the compact representations for the hop is used.
+    ///   0 -> sell and buy tokens remain unchanged from the previous fill (pure multiplex)
+    ///   1 -> sell token remains unchanged from the previous fill, buy token is read from `data` (diamond multiplex)
+    ///   2 -> sell token becomes the buy token from the previous fill, new buy token is read from `data` (multihop)
+    ///   3 -> both sell and buy token are read from `data`
+    ///
+    /// This function is also responsible for calling `_note`, which maintains the `notes` array and
+    /// the corresponding mapping in transient storage
+    function _getPoolKey(
```
dekz, line 273: nit: This does a lot more than just decodes a pool key, also switching the entire token context (and modifying `State`). `_processNextHop`?

duncancmt: You'll be horrified to know that the version of this function in #233 does _even more_. It's responsible for not only decoding the next hop from `data` and setting `state`, but also managing `notes`. Although it's much better abstracted and encapsulated. Perhaps this function does too many things? Maybe the logic from setting `zeroForOne` and after should be lifted back out into the loop in `unlockCallback` and this function should be named `_decodeHopTokens`?

src/core/UniswapV4.sol
```diff
@@ -0,0 +1,770 @@
...
+    ////
+    //// `state` exists to reduce stack pressure and to simplify and gas-optimize the process of
+    //// swapping. By keeping track of the sell token on each hop, we're able to compress the
+    //// representation of the fills required to satisfy the swap. Most often in a swap, the tokens
+    //// in adjacent fills are somewhat in common. By caching these tokens, we avoid having them
+    //// appear multiple times in the calldata. Additionally, this caching helps us avoid reading
+    //// the credit of each fill's sell token.
+
+    /// Because we have to ABIEncode the arguments to `.swap(...)` and copy the `hookData` from
+    /// calldata into memory, we save gas be deallocating at the end of this function.
```
dekz, line 234:
```suggestion
    /// calldata into memory, we save gas by deallocating at the end of this function.
```
nit: spelling

PR #253, opened 2024-10-18
Add a Safe Guard that implements a timelock for the upgrade Safe

duncancmt:


test/SafeGuard.t.sol
```diff
@@ -0,0 +1,918 @@
...
+
+        // Heck yeah, bubble sort
+        {
+            Vm.Wallet memory tmp;
+            for (uint256 i = 1; i < owners.length; i++) {
+                for (uint256 j = i; j > 0; j--) {
+                    if (owners[j - 1].addr > owners[j].addr) {
+                        tmp = owners[j - 1];
+                        owners[j - 1] = owners[j];
+                        owners[j] = tmp;
```
e1Ru1o, line 206: do you know if this will be any better?
```suggestion
                        (owners[j - 1], owners[j]) =  (owners[j], owners[j - 1]);
```

e1Ru1o (approved):

test/SafeGuard.t.sol
duncancmt: Yeah, this is better. I think I wrote it this way because the assignment semantics are a bit complicated. I think what you've written is equivalent, though https://docs.soliditylang.org/en/latest/types.html#data-location-and-assignment-behavior

duncancmt: lol
```
Warning (7238): This assignment performs two copies to storage. Since storage copies do not first copy to a temporary location, one of them might be overwritten before the second is executed and thus may have unexpected effects. It is safer to perform the copies separately or assign to storage pointers first.
   --> test/0.8.25/SafeGuard.t.sol:203:25:
    |
203 |                         (owners[j - 1], owners[j]) =  (owners[j], owners[j - 1]);
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

that's why. because these are both storage references, this ends up working the same as
```
owners[j-1] = owners[j];
owners[j] = owners[j-1];
```
which is clearly incorrect

PR #255, opened 2024-10-24
Fix Velodrome convergence edge case, add a unit test for convergence

duncancmt:


src/core/Velodrome.sol
```diff
@@ -26,87 +28,92 @@ interface IVelodromePair {
...
-                            // There's no closer answer than y + 1
-                            return y + 1;
-                        }
-                        dy = 1;
-                    }
+                    uint256 dy = (xy - k).unsafeMulDiv(_VELODROME_NEWTON_BASIS, d);
                     y += dy;
+                    if (dy <= _VELODROME_NEWTON_EPS) {
+                        return y + dy;
+                    }
```
alexfertel, line 112: I got a bit confused here: You say that once we are within this bound we're good, but you're adding `dy` to `y` again right after updating the `y` (Effectively being `y + 2 * dy`). Is this just in case or something else? I feel like I'm missing some context.

alexfertel (approved):
LGTM! Only one question for my edification.

test/integration/VelodromePairTest.t.sol
```diff
@@ -91,4 +239,48 @@ contract VelodromePairTest is BasePairTest {
...
+        vm.assume(max_dx >= _VELODROME_NEWTON_BASIS);
+        dx = bound(dx, _VELODROME_NEWTON_BASIS, max_dx);
+
+        uint256 k = dummy.k(x, y);
+        assertGe(k, _VELODROME_NEWTON_BASIS);
+
+        uint256 dy = dummy.get_y(x, dx, y);
+
+        //assertGe(dummy.k(x + dx, y - dy), k);
+        //assertLt(dummy.k(x + dx, y - dy - 2 * _VELODROME_NEWTON_EPS), k - _VELODROME_NEWTON_EPS);
```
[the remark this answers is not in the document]
duncancmt: This has been uncommented and the assertion now passes

src/core/Velodrome.sol
duncancmt: This was to account for rounding error in the computation of `k` in the AMM. I've replaced this code with something a bit more explicit and well-commented.

alexfertel (approved):

dekz (approved):

PR #280, opened 2025-01-10
[dependabot] add slack alerts

davidk-zx:
Enables dependabot slack alerts for `0x-settler`

@0xProject/platform-engineering


duncancmt:
Hi @davidk-zx please don't do that. Especially not when the CODEOWNERS kicks in. I'm gonna force-push `master` back to the state it was previously. Please re-open this PR

PR #285, opened 2025-01-14
Remove a signer from the deployment Safe

duncancmt:


sh/remove_signer.sh
```diff
@@ -0,0 +1,194 @@
...
+declare -r packed_signatures
+
+# configure gas limit
+declare -r -a args=(
+    "$safe_address" "$execTransaction_sig"
+    # to, value, data, operation, safeTxGas, baseGas, gasPrice, gasToken, refundReceiver, signatures
+    "$safe_address" 0 "$removeOwner_call" 0 0 0 0 "$(cast address-zero)" "$(cast address-zero)" "$packed_signatures"
+)
+
+# set gas limit and add multiplier/headroom (again mostly for Arbitrum)
```
kyu-c, line 184: probably need this for Mantle too

kyu-c (approved):

sh/remove_signer.sh
duncancmt: We merely double the gas limit on Mantle and set the gas price to 1 gwei. What do you think are reasonable values there?

duncancmt: Mantle is OP-stack so shouldn't the DA fee be billed directly in MNT rather than gas? And we do use `eth_estimateGas` to figure out the gas consumption, so the buffer is only there to compensate for errors in that RPC. Do we really need to 3kx the gas limit (not just the price) in order to get reasonable execution?

Unless you mean that we should increase the gas price substantially?

kyu-c: oh i see. if we are calling `eth_estimateGas` then it should be fine! 

Mantle deviated a lot from OP-stack and they charge DA fee explicitly (similar to how Arbitrum does).

PR #291, opened 2025-01-24
[dependabot] adjust slack alerts

davidk-zx:
- run upon every master branch push
- run every day

@0xProject/platform-engineering


duncancmt (commented):
This looks fine, but in the future, please wait for review before merging, especially if your PR triggers the `CODEOWNERS`. That's particularly relevant right now because the build and deploy system is particular about the `HEAD` from which it runs.

PR #307, opened 2025-02-20
Update MonadTestnet explorer configuration now that it's public

duncancmt:


wonge97 (approved):

wonge97:
i still have outstanding question to Monad team about which will be the providers of the official explorers

PR #308, opened 2025-02-25
Secrets encrypted at rest

dekz:


sh/common_secrets.sh
```diff
@@ -3,26 +3,49 @@ if ! hash sha256sum &>/dev/null ; then
...
+if [ -f "$project_root"/secrets.json ] ; then
+    echo 'secrets.json exists, remove it - will use secrets.json.scrypt only' >&2
     exit 1
 fi
 
+declare secrets_storage
+
+function decrypt_secrets {
+    local password
+    echo 'Enter passphrase for secrets.json.scrypt'
```
duncancmt, line 30: I don't think this is used anywhere? Is it?

sh/common_secrets.sh
```diff
@@ -3,26 +3,49 @@ if ! hash sha256sum &>/dev/null ; then
...
+    local decrypted
+    decrypted="$(scrypt dec "$project_root"/secrets.json.scrypt)"
+    if [ $? -ne 0 ]; then
```
duncancmt, line 33: the `set -e` that's enabled in every script should make a failure of `scrypt` cause the entire script to bail out before you even get to the `if`

sh/common_secrets.sh
```diff
@@ -3,26 +3,49 @@ if ! hash sha256sum &>/dev/null ; then
...
+    if ! echo "$decrypted" | sha256sum | grep -q "^bb82de121880f1182dbae410b341749e5ac1355954ae6c03151a1826e7bba745"; then
+        echo "Decrypted secrets.json hash verification failed" >&2
+        exit 1
+    fi
+    secrets_storage="$decrypted"
+}
+
 function get_secret {
-    jq -Mr ."$1"."$2" < "$project_root"/secrets.json
+    if [ -z "$secrets_storage" ]; then
```
duncancmt, line 47: The `set -eu` that's enabled in every script will cause the expansion of `"$secrets_storage"` to cause the whole script to bail out if it's unset. I recommend `"${secrets_storage-}" to suppress that behavior.

Also for hygiene, you might want to `declare secrets_storage` before this function

sh/common_secrets.sh
```diff
@@ -3,26 +3,49 @@ if ! hash sha256sum &>/dev/null ; then
...
+    echo 'Enter passphrase for secrets.json.scrypt'
+    local decrypted
+    decrypted="$(scrypt dec "$project_root"/secrets.json.scrypt)"
+    if [ $? -ne 0 ]; then
+        echo "Failed to decrypt secrets.json.scrypt" >&2
+        exit 1
+    fi
+
+    # 24290900be9575d1fb6349098b1c11615a2eac8091bc486bec6cf67239b7846a previous version prior to allowanceHolderLondon
+    if ! echo "$decrypted" | sha256sum | grep -q "^bb82de121880f1182dbae410b341749e5ac1355954ae6c03151a1826e7bba745"; then
```
duncancmt, line 39: I feel like there's gotta be a better way to do this. Using a heredoc `<<<` appends an extra newline to the contents of the variable, but maybe that's OK?

sh/common.sh
```diff
@@ -40,9 +40,15 @@ if [[ $api_secrets_permissions != '-rw-------' ]] ; then
...
 
+if [ -z ${1+x} ] ; then
+    echo 'chain_name argument is missing' >&2
+    exit 1
+fi
 declare -r chain_name="$1"
 shift
 
+
+
```
duncancmt, line 51: nit

PR #322, opened 2025-04-08
Ekubo VIP

e1Ru1o:


src/core/Ekubo.sol
```diff
@@ -0,0 +1,372 @@
...
+                assembly ("memory-safe") {
+                    sellToken := and(_ADDRESS_MASK, sellToken)
+                    buyToken := and(_ADDRESS_MASK, buyToken)
+                    zeroForOne :=
+                        or(
+                            eq(sellToken, 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee),
+                            and(iszero(eq(buyToken, 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee)), lt(sellToken, buyToken))
+                        )
+                }
+                (poolKey.token0, poolKey.token1) = zeroForOne.maybeSwap(address(sellToken), address(buyToken));
```
duncancmt, line 257: I think this is backwards

src/core/Ekubo.sol
```diff
@@ -0,0 +1,372 @@
...
+                data.offset := add(0x20, data.offset)
+                data.length := sub(data.length, 0x20)
+                // we don't check for array out-of-bounds here; we will check it later in `Decoder.overflowCheck`
+            }
+
+            Decoder.overflowCheck(data);
+
+            {
+                SqrtRatio sqrtRatio =
+                    SqrtRatio.wrap(uint96(zeroForOne.ternary(uint256(4611797791050542631), uint256(79227682466138141934206691491))));
```
duncancmt, line 283: :eyes: funky constants. I assume that these are different due to Ekubo's weird fixnum encoding?

src/core/Ekubo.sol
```diff
@@ -0,0 +1,372 @@
...
+                (int256 delta0, int256 delta1) =
+                    CORE.swap_611415377(poolKey, int128(amountSpecified), zeroForOne, sqrtRatio, skipAhead);
```
duncancmt, line 285: What should we do if `amountSpecified` overflows an `int128`?

src/core/Ekubo.sol
```diff
@@ -0,0 +1,372 @@
...
+                (int256 settledSellAmount, int256 settledBuyAmount) = zeroForOne.maybeSwap(delta0, delta1);
```
duncancmt, line 286: I think this is also backwards

src/core/Ekubo.sol
```diff
@@ -0,0 +1,372 @@
...
+        uint256 sellAmount,
+        ISignatureTransfer.PermitTransferFrom calldata permit,
+        bool isForwarded,
+        bytes calldata sig
+    ) private returns (uint256) {
+        if (sellToken == ETH_ADDRESS) {
+            // TODO: Check if this is a valid case
+            // No need to call pay, the receive function will be used to pay the debt
+            SafeTransferLib.safeTransferETH(payable(address(CORE)), sellAmount);
+            return sellAmount;
```
duncancmt, line 175: interesting. so this is part of the difference between UniV4 and Ekubo where payment isn't done using a `sync`/`settle` flow, but instead a callback?

src/core/Ekubo.sol
```diff
@@ -0,0 +1,372 @@
...
+        } else {
+            // Encode the call plus the extra data that is going to be needed in the callback
+            bytes memory data = abi.encodePacked(
+                abi.encodeCall(IEkuboCore.pay, address(sellToken)), abi.encode(payer, sellAmount, permit, isForwarded, sig)
+            );
+            bytes memory encodedPayedAmount = _setOperatorAndCall(
+                address(CORE), data, uint32(IEkuboCallbacks.payCallback.selector), payCallback
```
duncancmt, line 182:
```suggestion
                msg.sender, data, uint32(IEkuboCallbacks.payCallback.selector), payCallback
```
if this only is called inside the lock, then we can save some bytes by avoiding pushing the literal vault

duncancmt (commented):
This looks pretty good. I called out a couple things I don't understand as well as what I *think* are a couple bugs. It still need some plumbing and golfing before it can go in, but I like it :+1:

src/core/Ekubo.sol
duncancmt: oh also, 
```suggestion
                    IEkuboCore(msg.sender).swap_611415377(poolKey, int128(amountSpecified), zeroForOne, sqrtRatio, skipAhead);
```
save some bytes

src/core/Ekubo.sol
e1Ru1o: yep, they don't have `sync`/`settle`. You need to call `pay` specifying a token to repay and they will invoke `payCallback` to collect the payment. That is for tokens. For native eth you just need to send it directly

src/core/Ekubo.sol
e1Ru1o: Ok ok, wasn't sure here and decided to play safe. In other contracts msg.sender was a vault but here it is always CORE.

src/core/Ekubo.sol
e1Ru1o: Yep, taken from their contract
image

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+                data.length := sub(data.length, 0x02)
+                // we don't check for array out-of-bounds here; we will check it later in `Decoder.overflowCheck`
+            }
+
+            data = Decoder.updateState(state, notes, data);
+            // TODO: check the sign convention
+            int256 amountSpecified = int256((state.sell.amount * bps).unsafeDiv(BASIS)).unsafeNeg();
+            if (amountSpecified > type(int128).max) {
+                revert AmountSpecifiedTooLarge(amountSpecified);
+            }
```
duncancmt, line 330:
```suggestion
            if (amountSpecified >> 127 != amountSpecified >> 128) {
                Panic.panic(Panic.ARITHMETIC_OVERFLOW);
            }
```

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+                bps := shr(0xf0, calldataload(data.offset))
+
+                data.offset := add(0x02, data.offset)
+                data.length := sub(data.length, 0x02)
+                // we don't check for array out-of-bounds here; we will check it later in `Decoder.overflowCheck`
+            }
+
+            data = Decoder.updateState(state, notes, data);
+            // TODO: check the sign convention
+            int256 amountSpecified = int256((state.sell.amount * bps).unsafeDiv(BASIS)).unsafeNeg();
```
duncancmt, line 327: I'm not sure if this matters (or if it's handled correctly elsewhere) but we should either put this in an `unchecked` block or we should correctly handle overflow in `uint256 -> int256` conversion

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+        SqrtRatio sqrtRatioLimit,
+        uint256 skipAhead
+    ) internal returns (int128 delta0, int128 delta1) {
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+
+            // mstore(ptr, 0x00000000) // selector for `swap_611415377((address,address,bytes32),int128,bool,uint96,uint256)`
+            // TODO: Check that native ETH is handled properly (0x0 vs 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee)
+            mcopy(add(0x20, ptr), key, 0x60)
+            mstore(add(0x80, ptr), amount)
```
duncancmt, line 79: needs `SIGNEXTEND`

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+            mstore(add(0xa0, ptr), isToken1)
+            mstore(add(0xc0, ptr), sqrtRatioLimit)
```
duncancmt, line 81: Needs to be masked before putting in memory

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+                mstore(add(data, 0x140), sig.length)
+                calldatacopy(add(data, 0x160), sig.offset, sig.length)
+
+                let size := add(0x124, sig.length)
+
+                // update data length
+                mstore(data, size)
+
+                // update free memory pointer (word-aligned)
+                mstore(0x40, add(data, and(add(add(size, 32), 31), not(31))))
```
duncancmt, line 258: We don't care that the free memory pointer is word-aligned. Also combine constants where possible, use hex constants, and put constants on the left of commutative operators

src/core/Ekubo.sol
e1Ru1o: not going to miss this one next time, sorry

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+
+    function unsafeWithdraw(IEkuboCore core, address token, address recipient, uint128 amount) internal {
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+
+            mstore(ptr, 0x03a65ab6) // selector for `withdraw(address,address,uint128)`
+            // TODO: Check that native ETH is handled properly (0x0 vs 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee)
+            mstore(add(0x20, ptr), token)
+            mstore(add(0x40, ptr), recipient)
+            mstore(add(0x60, ptr), amount)
```
duncancmt, line 114: All these values need to be masked/cleaned before putting them in memory. You can probably get a bit clever with this (see how Solady's SafeTransferLib handles `transferFrom`)

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+    function payCallback(bytes calldata data) private returns (bytes memory) {
+        (
+            IERC20 sellToken,
+            uint256 requestedAmount,
+            address payer,
+            uint256 sellAmount,
+            ISignatureTransfer.PermitTransferFrom memory permit,
+            bool isForwarded,
+            bytes memory sig
+        ) = abi.decode(data, (IERC20, uint256, address, uint256, ISignatureTransfer.PermitTransferFrom, bool, bytes));
```
duncancmt, line 447: I wonder if it's possible to do some of this work in `_ekuboPay` so that we don't have to `abi.decode` the whole `PermitTransferFrom` object if `address(this)` is the payer

src/core/Ekubo.sol
```diff
@@ -0,0 +1,459 @@
...
+            sellToken,
+            bps,
+            feeOnTransfer,
+            hashMul,
+            hashMod,
+            fills,
+            amountOutMin
+        );
+        bytes memory encodedBuyAmount =
+            _setOperatorAndCall(address(CORE), data, uint32(IEkuboCallbacks.locked.selector), _ekuboLockCallback);
```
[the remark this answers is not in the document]
e1Ru1o: Sorry I am missing why this here should be msg.sender? 🤔 first parameter is meant to be the `target` and `CORE` is not the caller of `sellToEkubo` afaik

src/core/Ekubo.sol
duncancmt: I think there's similar logic in the UniV3 actions

src/core/Ekubo.sol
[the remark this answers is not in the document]
duncancmt: My bad. You're right. Got overzealous with the find/replace

src/core/Ekubo.sol
duncancmt: Actually, does this function need to exist at all? If `Take.take` is working correctly, we should not need it

src/core/Ekubo.sol
e1Ru1o: I kept this TODO from PancakeInfinity implementation. Seems like it was just removed so probably you already decided it was not relevant. See [here](https://github.com/0xProject/0x-settler/pull/311/commits/4ec119bc2e1da8d9caddf1da960f43326ce8d217)

duncancmt: Ahh, I'm not referring to the TODO (in PancakeInfinity, I removed it after double checking). I'm referring to casting from uint256 to int256. The upper half of uint256 values cannot be safely cast to int256 and Solidity does not throw an error when you do it

duncancmt: So we will revert on errors when the multiplcation overflows, but not when the typecast does. We should either verify that neither are possible (and put it in an `unchecked` block) or handle both by reverting

src/core/Ekubo.sol
```diff
@@ -254,8 +224,8 @@ abstract contract Ekubo is SettlerAbstract {
                 // update data length
                 mstore(data, size)
 
-                // update free memory pointer (word-aligned)
-                mstore(0x40, add(data, and(add(add(size, 32), 31), not(31))))
+                // update free memory pointer
+                mstore(0x40, add(0x20, add(data, size)))
```
duncancmt, line 228: Ahh, as a special case here, because `size := add(0x124, sig.length)`, it's actually faster/cheaper to write this as `add(data, add(0x20, size))` because the compiler will replace it with `add(data, add(0x144, sig.length))`

src/core/Ekubo.sol
e1Ru1o: yep, got the issue, but, do I need to double check again 🤔 ? Seems like it is exactly the same case as in PancakeInfinity, that is why I mentioned it

duncancmt: Given that all the amounts come from `int128` with the exception of the initial sell amount, I think we can use `unchecked` here. I think that we ought to check that during the header parsing we ought to check that `state.globalSell.amount` fits in an `int128` and then everything ought to be safe after that. What do you think?

src/core/Ekubo.sol
```diff
@@ -0,0 +1,453 @@
...
+        assert(BASIS == Decoder.BASIS);
+        assert(ETH_ADDRESS == Decoder.ETH_ADDRESS);
+    }
+
+    // fill encoding musings:
+    //     bps (2 bytes)
+    //     packing key (1 byte)
+    //     sell/buy tokens (0, 20, or 40 bytes)
+    //     config (32 bytes) - (8 bytes fee, 4 bytes tickSpacing, 20 bytes extension)
+    //     skipAhead (32 bytes)
```
duncancmt, line 114: There's no possibility for hookdata with Ekubo? It's not something clever where the hookdata is appended to the end of the call to `swap_611415377`?

src/core/Ekubo.sol
```diff
@@ -0,0 +1,453 @@
...
+            }
+        }
+    }
+
+    // the mandatory fields are
+    // 2 - sell bps
+    // 1 - pool key tokens case
+    // 32 - config (8 fee, 4 tickSpacing, 20 extension)
+    // 32 - skipAhead
+    uint256 private constant _HOP_DATA_LENGTH = 62;
```
duncancmt, line 248: I think this constant is wrong

src/core/Ekubo.sol
```diff
@@ -0,0 +1,453 @@
...
+        uint256 sellAmount;
+        
+        ISignatureTransfer.PermitTransferFrom calldata permit;
+        bool isForwarded;
+        bytes calldata sig;
+
+        
+        assembly ("memory-safe") {
+            
+        }
```
duncancmt, line 422: ?

src/core/Ekubo.sol
```diff
@@ -0,0 +1,453 @@
...
+            sellAmount := calldataload(add(0x40, data.offset))
+
+            if gt(calldatasize(), 0x60) {
+                permit := add(0x60, data.offset)
+                isForwarded := calldataload(add(0xc0, data.offset))
+                sig.length := calldataload(add(0xe0, data.offset))
+                sig.offset := add(0x100, data.offset)
+            }
+        }
+        // TODO: Check if  assert(requestedAmount == sellAmount) should be true
```
duncancmt, line 441: This is almost certainly not the case. With this kind of AMM `requestedAmount <= sellAmount` because the AMM may run out of liquidity.

src/core/Ekubo.sol
```diff
@@ -0,0 +1,453 @@
...
+            mstore(add(0xa0, ptr), isToken1)
+            mstore(add(0xc0, ptr), and(0xffffffffffffffffffffffff, sqrtRatioLimit))
+            mstore(add(0xe0, ptr), skipAhead)
+            
+            if iszero(call(gas(), core, 0x00, add(0x1c, ptr), 0xe4, 0x00, 0x40)) {
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
+            delta0 := mload(0x00)
+            delta1 := mload(0x20)
```
duncancmt, line 87: consider adding a comment about how you verified that `CORE` is well-behaved so we don't need to check whether there are dirty bits in the returndata (presuming that you did actually verify that fact)

src/core/Ekubo.sol
```diff
@@ -0,0 +1,453 @@
...
+                // update free memory pointer
+                mstore(0x40, add(data, add(0x20, size)))
+            }
+            bytes memory encodedPayedAmount =
+                _setOperatorAndCall(msg.sender, data, uint32(IEkuboCallbacks.payCallback.selector), payCallback);
+            assembly ("memory-safe") {
+                // We can skip all the checks performed by `abi.decode` because we know that this is the
+                // verbatim result from `payCallback` and that `payCallback` encoded the payment
+                // correctly.
+                payment := mload(add(0x60, encodedPayedAmount))
```
duncancmt, line 238: Ditto other comment about verifying whether there are dirty bits or skipping that if we know dirty bits are impossible

src/core/Ekubo.sol
```diff
@@ -0,0 +1,453 @@
...
+            uint32(IEkuboCore.lock.selector),
+            recipient,
+            sellToken,
+            bps,
+            feeOnTransfer,
+            hashMul,
+            hashMod,
+            fills,
+            amountOutMin
+        );
```
duncancmt, line 139: You should verify that this produces the expected result even though `lock` takes an `uint256` argument instead of a `bytes` argument. I think that just means that `id` will always be 32. And an integration test will tell you for sure. But it's worth thinking over to make sure that there's nothing unexpected

duncancmt (commented):
Looking polished :+1: 

Resolve your TODOs and add an integration test

src/core/Ekubo.sol
e1Ru1o: last minute push before going off 🤦‍♂️, moved the code into the next assembly section and did not remove that

src/core/Ekubo.sol
e1Ru1o: Thought so as well, will make sure of using requestedAmount

src/core/Ekubo.sol
e1Ru1o: Extensions (which are the "hooks" in Ekubo) don't receive any special data from the call, just some of the known parameters/data in use. So, it doesn't seem to be possible to have any hookdata.
image
image
Extensions docs [here](https://docs.ekubo.org/integration-guides/extensions)

src/core/Ekubo.sol
e1Ru1o: Yep, I was going to ask you about this one today but forgot, now is clear to me that it is just the amount of bytes mentioned in the comment 😅, updating it

src/core/Ekubo.sol
e1Ru1o: in this case lock takes no parameter, but got your point, double checking

duncancmt: Ahh, I got confused about the callback ABI and the CORE ABI. So we can safely ignore the `id` argument to both of the callback functions?

e1Ru1o: we either ignore them or use them for validation, we should receive the same id in all the callbacks, right now I am not doing any checks on that.

src/core/Ekubo.sol
```diff
@@ -311,6 +312,7 @@ abstract contract Ekubo is SettlerAbstract {
                             eq(sellToken, 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee),
                             and(iszero(eq(buyToken, 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee)), lt(sellToken, buyToken))
                         )
+                    isToken1 := iszero(zeroForOne)
```
duncancmt, line 315: you can probably golf the expression for `zeroForOne` so that you directly compute `isToken1` instead

src/core/Ekubo.sol
```diff
@@ -350,11 +354,12 @@ abstract contract Ekubo is SettlerAbstract {
...
                 // TODO: verify that the vault enforces that the settled sell amount cannot be
                 // positive
-                state.sell.amount -= uint256(settledSellAmount.unsafeNeg());
+                state.sell.amount -= uint256(settledSellAmount);
+
                 // If `state.buy.amount()` overflows an `int128`, we'll get a revert inside the
                 // vault later. We cannot overflow a `uint256`.
                 unchecked {
-                    state.buy.amount += settledBuyAmount.asCredit(state.buy.token);
+                    state.buy.amount += settledBuyAmount.unsafeNeg().asCredit(state.buy.token);
```
duncancmt, line 362: I think you should use `asDebt` instead. Of course, add a comment that the sign convention here is backwards from UniV4/BalV3/PancakeInfinity

test/integration/BalancerV3.t.sol
```diff
@@ -133,7 +133,7 @@ abstract contract BalancerV3Test is SettlerMetaTxnPairTest, AllowanceHolderPairT
     function testBalancerV3() public skipIf(balancerV3Pool() == address(0)) setBalancerV3Block {
         (ISignatureTransfer.PermitTransferFrom memory permit, bytes memory sig) = _getDefaultFromPermit2();
 
-        (uint256 hashMul, uint256 hashMod) = perfectHash();
+        (uint256 hashMul, uint256 hashMod) = BalancerV3Test.perfectHash();
```
duncancmt, line 136: what made this change necessary?

src/ISettlerActions.sol
```diff
@@ -202,4 +202,26 @@ interface ISettlerActions {
...
+    function EKUBO_VIP(
+        address recipient,
+        bool feeOnTransfer,
+        uint256 hashMul,
+        uint256 hashMod,
+        bytes memory fills,
+        ISignatureTransfer.PermitTransferFrom memory permit,
+        bytes memory sig,
+        uint256 amountOutMin
+    ) external;
```
duncancmt, line 226: There also needs to be a metatx variant (without the `sig` argument)

src/core/Ekubo.sol
```diff
@@ -72,13 +72,15 @@ library UnsafeEkuboCore {
             let ptr := mload(0x40)
 
             mstore(ptr, 0x00000000) // selector for `swap_611415377((address,address,bytes32),int128,bool,uint96,uint256)`
-            // TODO: Check that native ETH is handled properly (0x0 vs 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee)
-            mcopy(add(0x20, ptr), poolKey, 0x60)
+            let token0 := mload(poolKey)
+            token0 := mul(token0, iszero(eq(0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, token0)))
+            mstore(add(0x20, ptr), token0)
+            mcopy(add(0x40, ptr), add(0x20, poolKey), 0x40)
```
duncancmt, line 78: You should check this, but I think you may save on contract size if you
```
let firstArg := add(0x20, ptr)
mcopy(firstArg, poolKey, 0x60)
let token0 := mload(firstArg)
mstore(firstArg, mul(iszero(eq(0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, token0)), token0))
```

src/core/Ekubo.sol
```diff
@@ -295,13 +296,13 @@ abstract contract Ekubo is SettlerAbstract {
             }
 
             data = Decoder.updateState(state, notes, data);
-            // TODO: check the sign convention
-            int256 amountSpecified = int256((state.sell.amount * bps).unsafeDiv(BASIS)).unsafeNeg();
+            int256 amountSpecified = int256((state.sell.amount * bps).unsafeDiv(BASIS));
             if (amountSpecified >> 127 != amountSpecified >> 128) {
                 Panic.panic(Panic.ARITHMETIC_OVERFLOW);
             }
```
duncancmt, line 302: This can be simplified now that `amountSpecified` is positive

src/core/Ekubo.sol
e1Ru1o: it is not guaranteed, masking them ...

src/core/Ekubo.sol
e1Ru1o: the comment is there already, ecodedPayment came from payCallback which is our function and encodes data properly

test/integration/BalancerV3.t.sol
e1Ru1o: This is because I kept the `perfectHash` with the same name in both EkuboTest and BalancerV3Test, so USDCUSDTTest needs to override `perfectHash` then I needed to use the explicit one I want to use on each, maybe better to just change the names of the functions to avoid the override (which is not needed). I am going to do that

src/core/Ekubo.sol
e1Ru1o: You are right
```
╭---------------------------------+------------------+-------------------+--------------------+---------------------╮
| Contract                        | Runtime Size (B) | Initcode Size (B) | Runtime Margin (B) | Initcode Margin (B) |
+===================================================================================================================+
| MainnetSettler   (Original)     | 21,803           | 23,361            | 2,773              | 25,791              |
| MainnetSettler   (   New  )     | 21,785           | 23,343            | 2,791              | 25,809              |
╰---------------------------------+------------------+-------------------+--------------------+---------------------╯
```

src/core/Ekubo.sol
e1Ru1o: Sounds good doing so.

test/integration/USDC_WETH.t.sol
```diff
@@ -128,8 +131,36 @@ contract USDCWETHTest is
...
+    }
+
+    function ekuboExtraActions(bytes[] memory actions) internal view virtual override returns (bytes[] memory) {
+        bytes[] memory data = new bytes[](actions.length + 2);
+        address _weth = address(toToken());
+        for (uint256 i; i < actions.length; i++) {
+            data[i] = actions[i];
+        }
+        data[actions.length] = abi.encodeCall(ISettlerActions.BASIC, (_eth, 10_000, address(_weth), 0, ""));
+        data[actions.length + 1] = abi.encodeCall(ISettlerActions.BASIC, (_weth, 10_000, address(_weth), 36, abi.encodeCall(toToken().transfer, (FROM, uint256(0)))));
```
duncancmt, line 163: I think you can use the slippage limit to perform this transfer, unless that's too much of a pain

e1Ru1o: Is that relevant to be tested in this PR? I would prefer to keep it as it is to avoid adding more differences between USDC<>USDT and USDC<>WETH.

src/core/Ekubo.sol
```diff
@@ -0,0 +1,486 @@
...
+            mstore(add(0xa0, ptr), isToken1)
+            mstore(add(0xc0, ptr), and(0xffffffffffffffffffffffff, sqrtRatioLimit))
+            mstore(add(0xe0, ptr), skipAhead)
+
+            if iszero(call(gas(), core, 0x00, add(0x1c, ptr), 0xe4, 0x00, 0x40)) {
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
+            delta0 := and(0xffffffffffffffffffffffffffffffff, mload(0x00))
+            delta1 := and(0xffffffffffffffffffffffffffffffff, mload(0x20))
```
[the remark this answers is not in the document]
duncancmt: Of course, if we can know that `delta0` and `delta1` never have dirty bits, we can skip the whole exercise

src/core/Ekubo.sol
```diff
@@ -0,0 +1,486 @@
...
+    //// i.e. encode only the buy token for the first fill.
+    ////   0 -> sell and buy tokens remain unchanged from the previous fill (pure multiplex)
+    ////   1 -> sell token remains unchanged from the previous fill, buy token is encoded (diamond multiplex)
+    ////   2 -> sell token becomes the buy token from the previous fill, new buy token is encoded (multihop)
+    ////   3 -> both sell and buy token are encoded
+    //// Obviously, after encoding the packing key, you encode 0, 1, or 2 tokens (each as 20 bytes),
+    //// as appropriate.
+    //// The remaining fields of the fill are mandatory.
+    //// Third, encode the config of the pool as 32 bytes
+    //// Fourth, encode the skipAhead to use in the swap as 32 bytes
```
duncancmt, line 139: Maybe explain the semantics of these fields by copying Ekubo's documentation? Or at least link to Ekubo's documentation

src/core/Ekubo.sol
```diff
@@ -0,0 +1,486 @@
...
+
+                data.offset := add(0x02, data.offset)
+                data.length := sub(data.length, 0x02)
+                // we don't check for array out-of-bounds here; we will check it later in `Decoder.overflowCheck`
+            }
+
+            data = Decoder.updateState(state, notes, data);
+            int256 amountSpecified;
+            unchecked {
+                amountSpecified = int256((state.sell.amount * bps).unsafeDiv(BASIS));
```
duncancmt, line 330: It might be good to briefly note why it's not possible for this operation to overflow

src/core/Ekubo.sol
```diff
@@ -0,0 +1,486 @@
...
+                    poolKey, int128(amountSpecified), isToken1, sqrtRatio, skipAhead
+                );
+                // Ekubo sign convention here is backwards compared to UniV4/BalV3/PancakeInfinity
+                // settledSellAmount is positive, settledBuyAmount is negative
+                (int256 settledSellAmount, int256 settledBuyAmount) = isToken1.maybeSwap(delta0, delta1);
+
+                state.sell.amount -= settledSellAmount.asCredit(state.sell.token);
+
+                // If `state.buy.amount()` overflows an `int128`, we'll get a revert inside the
+                // vault later. We cannot overflow a `uint256`.
```
duncancmt, line 382: Is this true? What will happen if we overflow an `int128`? I wonder if it might be better to leave the argument to `unsafeSwap` as `int256` (omitting the `SIGNEXTEND` before writing the value to memory) and then let the ABI decoding in `CORE` serve as our guard against overflow.

duncancmt (approved):

src/core/Ekubo.sol
```diff
@@ -0,0 +1,486 @@
...
+            SafeTransferLib.safeTransferETH(payable(msg.sender), sellAmount);
+            return sellAmount;
+        } else {
+            // Encode the call plus the extra data that is going to be needed in the callback
+            bytes memory data;
+            assembly ("memory-safe") {
+                data := mload(0x40)
+
+                mstore(add(0x04, data), 0x0c11dedd) // selector for pay(address)
+                mstore(add(0x24, data), and(_ADDRESS_MASK, sellToken))
```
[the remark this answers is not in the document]
e1Ru1o: No it is not going to mess anything I think, length of data is going to be the same

src/core/Ekubo.sol
e1Ru1o: Ekubo's doc is not that good to have a section explaining them but let me add comments

src/core/Ekubo.sol
```diff
@@ -0,0 +1,486 @@
...
+                // starts at the beginning of sellToken
+                permit := add(0x20, data.offset)
+                isForwarded := calldataload(add(0xa0, data.offset))
+                
+                sig.offset := add(0xc0, data.offset)
+                sig.length := calldataload(sig.offset)
+                sig.offset := add(0x20, sig.offset)
+            }
+        }
+        if (data.length == 0x60) {
```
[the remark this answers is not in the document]
e1Ru1o: this one actually broke the tests, seems harmless but the `<` case should go to the else, so it doesn't work here. We can still keep it like that and switch the if and else bodies.

src/core/Ekubo.sol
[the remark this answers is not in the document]
e1Ru1o: Skipping checks, we can assume not dirty bits, double checked again

PR #324, opened 2025-04-14
Add limit order fee collector contract

duncancmt:


test/integration/CurveTricryptoPairTest.t.sol
```diff
@@ -10,6 +10,7 @@ import {Settler} from "src/Settler.sol";
 import {SettlerBase} from "src/SettlerBase.sol";
 import {ISettlerActions} from "src/ISettlerActions.sol";
 import {ISignatureTransfer} from "@permit2/interfaces/ISignatureTransfer.sol";
+import {ISettlerBase} from "src/interfaces/ISettlerBase.sol";
```
e1Ru1o, line 13: This was added but `SettlerBase` import was not removed. It can be removed in most of the modified files inside `test/integration`

src/LimitOrderFeeCollector.sol
```diff
@@ -0,0 +1,459 @@
...
+
+type MakerTraits is uint256;
+
+/**
+ * @notice Checks if the maker needs to unwraps WETH.
+ * @param makerTraits The traits of the maker.
+ * @return result A boolean indicating whether the maker needs to unwrap WETH.
+ */
+function unwrapWeth(MakerTraits makerTraits) pure returns (bool) {
+    return MakerTraits.unwrap(makerTraits) & 0x80000000000000000000000000000000000000000000000000000000000000 != 0;
```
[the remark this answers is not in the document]
duncancmt: Ahh, this was actually reference code. TBQH, the way that this should _actually_ be implemented is with `>>`.

src/LimitOrderFeeCollector.sol
```diff
@@ -0,0 +1,459 @@
...
+    error ApproveFailed(IERC20 token);
+
+    constructor(bytes20 gitCommit, address initialOwner, address initialFeeCollector, IERC20 weth_) {
+        emit GitCommit(gitCommit);
+        require(initialOwner != address(0));
+        _setPendingOwner(initialOwner);
+        require(initialFeeCollector != address(0));
+        feeCollector = initialFeeCollector;
+        emit SetFeeCollector(feeCollector);
+        weth = weth_;
```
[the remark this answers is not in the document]
duncancmt: I don't think I follow why you'd want to reorder this. Can you elaborate?

Also revert reasons are ignored in the constructor. You can't bubble any kind of revert reason from initcode, so these are equivalent.

e1Ru1o (commented):
It looks quite good until now. `_swap` function is beautiful.

test/integration/LimitOrderFeeCollector.t.sol
```diff
@@ -0,0 +1,413 @@
...
+        swaps[1].recipient = payable(address(this));
+        swaps[1].sellToken = USDT;
+        swaps[1].buyToken = ETH;
+        swaps[1].minBuyAmount = 1 wei;
+        swaps[1].actions = actions1;
+        swaps[1].zid = bytes32(0);
+
+        uint256 beforeBalance = address(this).balance;
+
+        vm.expectEmit(false, true, true, false, address(USDC));
```
DennisSimon, line 376: We don't need to explicitly check for the second swap as its covered in the balance check, correct?

DennisSimon (approved):
Great job!
I worked through the main contract and (apart from some parts of the assembly that i'm honestly not familiar with), the logic looks fine to me!
The tests also look sound and in my opinion everything is covered.

test/integration/LimitOrderFeeCollector.t.sol
duncancmt: There's no event for transferring ETH from Settler to the recipient (the test contract). I guess I could check for the event that transfers WETH from the pair to Settler or for the WETH `Withdraw` event. But yeah, it's covered by the check that the ETH balance of the test contract increases over the call to `multiSwap`. Plus I checked the trace by running the test suite with `-vvvv`

duncancmt:
@DennisSimon what do you think of the idea of removing the `recipient` field of the `Swap` struct and move that argument out into the `multiSwap` function? Basically, reduce the size of calldata and reduce the gas cost of execution as well at the cost of making it no longer possible to swap to multiple different recipients. I could even go one step further and do the same to the `buyToken` argument.

DennisSimon:
@duncancmt , that sounds fine to me. The recipient would be the final receiver of the converted funds, correct? This would likely always be the same address (treasury wallet) so i don't think it's necessary to provide different recipients per swap. 
Same is applicable to the buyToken, as we likely always want to swap to the same asset.

PR #327, opened 2025-04-17
Integration test for PancakeInfinity

e1Ru1o:


src/core/PancakeInfinity.sol
```diff
@@ -80,14 +80,7 @@ interface IPancakeInfinityPoolManager {
...
         view
-        returns (
-            IERC20 currency0,
-            IERC20 currency1,
-            IHooks hooks,
-            IPancakeInfinityPoolManager poolManager,
-            uint24 fee,
-            bytes32 parameters
-        );
+        returns (PoolKey memory);
```
duncancmt, line 83: Are you sure about this?

src/core/PancakeInfinity.sol
```diff
@@ -277,6 +270,7 @@ abstract contract PancakeInfinity is SettlerAbstract {
     //// Obviously, after encoding the packing key, you encode 0, 1, or 2 tokens (each as 20 bytes),
     //// as appropriate.
     //// The remaining fields of the fill are mandatory.
+
```
duncancmt, line 273: ?

test/integration/BaseForkTest.t.sol
```diff
@@ -0,0 +1,17 @@
...
+
+contract MainnetDefaultFork is BaseForkTest {
+    function testChainId() internal pure virtual override returns (string memory) {
+        return "mainnet";
+    }
+    
+    function testBlockNumber() internal pure virtual override returns (uint256) {
+        return 18685612;
+    }
+}
```
duncancmt, line 17: wow! great abstraction!

duncancmt (approved):
Nice! I'll probably come back tomorrow morning and re-review, but 10pm Duncan thinks that this is good

src/core/PancakeInfinity.sol
e1Ru1o: yep, we can do so for simplicity

src/core/PancakeInfinity.sol
e1Ru1o: 😅, removing it

src/core/PancakeInfinity.sol
duncancmt: I agree that these are ABI-compatible, but what I'm saying is that it's probably preferable to do it the first way because it doesn't allocate. also it gives the compiler the opportunity to avoid loading values that we don't use

e1Ru1o: Fair enough but, maybe we should just get rid of that interface. It is not used in the contract at the moment is just used in the tests that is why I decided to change it 😅. I can remove it there and move to the test. In the tests the above arguments are not that relevant

PR #328, opened 2025-04-18
WIP: EIP7702 swap wallet

duncancmt:


src/eip7702/ZeroExEIP7702Wallet.sol
```diff
@@ -0,0 +1,250 @@
...
+    address private immutable _cachedThis;
+    string public constant name = "ZeroExEIP7702Wallet";
+    uint256 private immutable _cachedChainId;
+    bytes32 private immutable _cachedDomainSeparator;
+
+    error DeploymentFailed();
+    error PermissionDenied();
+
+    modifier onlyWallet() {
+        require(address(this).code.length == 23);
```
e1Ru1o, line 29:
```suggestion
        require(address(this).code.length == 38);
```

duncancmt: This is supposed to run in 7702'd EOAs that have exactly 23 bytes of code, beginning with 0xef0100

src/SettlerSwapper.sol
```diff
@@ -0,0 +1,237 @@
...
+            }
+        }
+    }
+
+    modifier validSettler(ISettlerTakerSubmitted settler) {
+        requireValidSettler(settler);
+        _;
+    }
+
+    function _swapAll(
```
e1Ru1o, line 77: wondering why do you prefer here to duplicate the entire function instead of just passing computing the amount and reusing the `_swap`. Is it the case that it is a huge improvement?

src/SettlerSwapper.sol
```diff
@@ -0,0 +1,237 @@
...
+                mstore(0x00, 0x70a08231) // selector for `balanceOf(address)`
+                mstore(0x20, address())
+                if iszero(staticcall(gas(), sellToken, 0x1c, 0x24, 0x40, 0x20)) { bubbleRevert(ptr) }
+                if iszero(lt(0x1f, returndatasize())) { emptyRevert() }
+
+                // Set the exact allowance on AllowanceHolder. The amount is already in memory 0x40.
+                mstore(0x00, 0x095ea7b3) // selector for `approve(address,uint256)`
+                mstore(0x20, ALLOWANCE_HOLDER_ADDRESS)
+                if iszero(call(gas(), sellToken, 0x00, 0x1c, 0x44, 0x00, 0x20)) { bubbleRevert(ptr) }
+                if iszero(or(and(eq(mload(0x00), 0x01), lt(0x1f, returndatasize())), iszero(returndatasize()))) {
```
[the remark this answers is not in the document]
e1Ru1o: if accepted, same on swap

src/eip7702/ZeroExEIP7702Wallet.sol
```diff
@@ -0,0 +1,250 @@
...
+                == keccak256(
+                    "Swap(uint256 nonce,address settler,address sellToken,uint256 sellAmount,address recipient,address buyToken,uint256 minAmountOut,bytes actions,bytes32 zid)"
+                )
+        );
+        uint256 $int;
+        Storage storage $ = _$();
+        assembly ("memory-safe") {
+            $int := $.slot
+        }
+        require($int == (uint256(_NAMEHASH) - 1) & 0xffffffffffffffffffffff00);
```
e1Ru1o, line 52: why this specific mask instead of ~0xff? isn't this a bit agains the standard?

src/SettlerSwapper.sol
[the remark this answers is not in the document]
duncancmt: because this entire line is constants/immediates, we don't reduce the stack pressure by moving variable references to the right. I can make this change for consistency, though

PR #331, opened 2025-04-21
WIP: dirty hack single-signature witness-bearing allowances

duncancmt:


src/interfaces/IERC5267.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+interface IERC5267 {
+    /// @notice Returns the un-hashed fields that comprise the EIP-712 domain.
+    /// @return fields A bitmask indicating which domain fields are used.
+    /// @return name The human-readable name of the domain.
+    /// @return chainId The chain ID of the network.
```
e1Ru1o, line 8: If adding the natspec, why incomplete?

e1Ru1o: and more or less same question for the named return parameters

src/utils/TransactionEncoder.sol
```diff
@@ -0,0 +1,135 @@
...
+    function encode(AccessListElem[] memory accessList) internal pure returns (LibRLP.List memory list) {
+        uint256 accountsLength = accessList.length;
+        for (uint256 i; i < accountsLength; i = i.unsafeInc()) {
+            AccessListElem memory elem = accessList[i];
+            LibRLP.List memory slots;
+            uint256 slotsLength = elem.slots.length;
+            for (uint256 j; j < slotsLength; j = j.unsafeInc()) {
+                slots.p(bytes.concat(elem.slots[j]));
+            }
+            list.p(LibRLP.p().p(elem.account).p(slots));
```
e1Ru1o, line 27: I think this can be removed, is it needed?
```suggestion
            list.p(LibRLP.p(elem.account).p(slots));
```

src/SingleSignatureDirtyHack.sol
```diff
@@ -0,0 +1,272 @@
...
+    // While the check on the allowance *SHOULD* be sufficient to prevent replay, we duplicate the
+    // protocol-level nonce at the application level to guard against weird ERC20s that might not
+    // decrease `allowance`. ERC20 does not require that `transferFrom` decrease the existing
+    // allowance.
+    function _consumeNonce(address owner, uint256 incomingNonce) private {
+        uint256 currentNonce = nonces[owner];
+        nonces[owner] = incomingNonce;
+        if (incomingNonce <= currentNonce) {
+            revert NonceReplay(currentNonce, incomingNonce);
+        }
```
e1Ru1o, line 126:
```suggestion
        if (incomingNonce <= currentNonce) {
            revert NonceReplay(currentNonce, incomingNonce);
        }
        nonces[owner] = incomingNonce;
```

duncancmt: Solc is dumb and is only able to optimize away (sometimes, when it feels like it) the duplicate `nonces[owner]` hash in the former case. Since the latter only optimizes the reverting path, I think it's preferable to have it the way that it is now. Unless you're thinking of something else?

PR #338, opened 2025-04-30
CrossChainReceiverFactory contract 

e1Ru1o:
See followup PR #352 for some helpful comments in `_verifyERC7739NestedTypedSignature`

---

Notes for auditors:

This contract is a factory for minimal proxies for wallets that receive funds from bridges. These wallets have, baked-in, a Merkle tree of preapproved hashes; the Merkle root is used to derive the `CREATE2` salt. This allows for cross-chain swaps to happen in a single user interaction because the destination address of the tokens implicitly encodes the swap on that chain. The idea is that the funds arrive in the counterfactual address where the minimal proxy will be deployed, then to execute the swap you deploy the minimal proxy (deterministically), approve the token to Permit2, and execute the swap using the witness feature of the Permit2 `SignatureTransfer`. The mechanism for the swap is irrelevant for the scope of this audit, but it relies on SettlerIntent, also in the same repository. Permit2 handles all the nonce cancelation, so the minimal proxy can be (nearly) stateless. There's also ERC7739 support for fallback flows as well as a generic `call` function for flows that can't be fully supported gaslessly.

Additional complexity is introduced for the sake of gas optimization and for good native token support (without changing the factory address). Using `selfdestruct` instead of storing the owner address upon proxy construction saves ~10k gas. And in order to have native token support, we have to be able to wrap ETH to WETH and approve it to Permit2, which would ordinarily make the factory's deployment address depend on the address of WETH, which changes across chains, so there's a convoluted system that allows that to be injected into the runtime bytecode of the factory without having it hashed.

---

Questions for auditors:

* Is there a scheme for nested, rehashed (ERC7739) smart contract (ERC1271) signatures? In other words, how can we make the signature scheme work when `owner()` is itself a smart contract?
* Is this compatible with ERC6492?
* This contract has been written for chains supporting Shanghai or higher hardfork. Are there noteworthy chains that require London support, once Linea has upgraded to (roughly) Prague?
* Can you think of a good reason why we ought to expose `DOMAIN_SEPARATOR()` in addition to the ERC5267-mandated fields?


src/BridgeFactory.sol
```diff
@@ -0,0 +1,117 @@
...
+        virtual
+        override
+        returns (bytes calldata)
+    {
+        return msg.data;
+    }
+
+    function _isForwarded() internal view virtual override returns (bool) {
+        return false;
+    }
```
duncancmt, line 116: Do we need this here, or can we inherit from `Context`?

src/BridgeFactory.sol
```diff
@@ -0,0 +1,117 @@
...
+        bytes32 root = MerkleProofLib.verify(proof, _hash);
+
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+
+            mstore(0x00, 0x60265f8160095f39f35f5f365f5f37365f6c)
+            extcodecopy(address(), 0x20, 0x09, 0x1d)
+            let initCodeHash := keccak256(0x0e, 0x2f)
+
+            mstore(0x00, 0xff00000000000000) // 0xff with padding for factory address
```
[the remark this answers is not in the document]
duncancmt: oh, I see. you've done a clever thing with the `extcodecopy` above. the factory address is already in memory at the correct offset

src/BridgeFactory.sol
e1Ru1o: yeah, surely not needed, added it just to push

duncancmt:
Some more idle thoughts that I had:
* we should mix the chainid into the salt when doing the create to avoid cross-chain replay (even considering that a "reasonable" ERC1271 requester should do this too)
* we should encode the authority that can execute the bail-out flow somewhere

src/BridgeFactory.sol
```diff
@@ -0,0 +1,101 @@
...
+pragma solidity ^0.8.25;
+
+import {IERC1271} from "./interfaces/IERC1271.sol";
+import {IERC20} from "@forge-std/interfaces/IERC20.sol";
+import {TwoStepOwnable} from "./deployer/TwoStepOwnable.sol";
+import {SafeTransferLib} from "./vendor/SafeTransferLib.sol";
+import {MerkleProofLib} from "./vendor/MerkleProofLib.sol";
+import {Context} from "./Context.sol";
+
+contract BridgeFactory is IERC1271, TwoStepOwnable, Context {
```
duncancmt, line 11: `Context` always has to go first after interfaces. You get some _very_ strange and wrong behavior if it comes late in the MRO

src/BridgeFactory.sol
```diff
@@ -0,0 +1,101 @@
...
+            // restore clobbered slots
+            mstore(0x60, 0x00)
+            mstore(0x40, ptr)
+
+            if shl(0x60, xor(computedAddress, address())) {
+                return(0x60, 0x20)
+            }
+            // Return ERC1271 magic value (isValidSignature selector)
+            mstore(0x00, 0x1626ba7e00000000000000000000000000000000000000000000000000000000)
+            return(0x00, 0x20)
```
duncancmt, line 70: since this block never returns to Solidity, we can optimize
```suggestion
            if shl(0x60, xor(computedAddress, address())) {
                mstore(0x00, 0x00)
                return(0x00, 0x20)
            }

            // Return ERC1271 magic value (isValidSignature selector)
            mstore(0x00, 0x1626ba7e00000000000000000000000000000000000000000000000000000000)
            return(0x00, 0x20)
```

src/BridgeFactory.sol
```diff
@@ -0,0 +1,101 @@
...
+    constructor() {
+        require(
+            (msg.sender == 0x4e59b44847b379578588920cA78FbF26c0B4956C &&
+                uint160(address(this)) >> 104 == 0) || block.chainid == 31337
+        );
+        _cachedThis = address(this);
+    }
+
+    modifier onlyWallet() {
+        require(address(this).code.length == 38);
```
duncancmt, line 25: The snippet that you commented on over in #328 is kind of inefficient, but that's the most expedient way to detect 7702'd wallets. This modifier, on the other hand, could probably be simpler because it just has to detect proxies:
```suggestion
        require(address(this) != _cachedThis);
```

duncancmt: I guess you could also name the modifier `onlyProxy` because that's a bit more specific

src/BridgeFactory.sol
e1Ru1o: yep, actually noticed while debugging that 0x60 and 0x40 were being skipped and I was wondering about removing them, now this is a confirmation

e1Ru1o:
Ready to merge after merging #340

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,332 @@
...
+            proof.offset := add(0x20, proof.offset)
+        }
+
+        _verifyRoot(MerkleProofLib.getRoot(proof, hash), owner);
+        return IERC1271.isValidSignature.selector;
+    }
+
+    function eip712Domain()
+        external
+        view
```
duncancmt, line 168: needs `onlyProxy`

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,332 @@
...
+            mstore(0x40, ptr)
+            mstore(0x60, 0x00)
+        }
+    }
+
+    function _DOMAIN_SEPARATOR() internal view returns (bytes32) {
+        return block.chainid == _cachedChainId ? _cachedDomainSeparator : _computeDomainSeparator();
+    }
+
+    function DOMAIN_SEPARATOR() external view noDelegateCall returns (bytes32) {
```
duncancmt, line 106: Should this be `noDelegateCall` or `onlyProxy`?

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,644 @@
...
+                    if or(eq(hash, b), lt(gas(), gasToBurn)) { invalid() }
+                    // Make a call to this with `b`, efficiently burning the gas provided.
+                    // No valid transaction can consume more than the gaslimit.
+                    // See: https://ethereum.github.io/yellowpaper/paper.pdf
+                    // Most RPCs perform calls with a gas budget greater than the gaslimit.
+                    pop(staticcall(gasToBurn, address(), 0x1c, 0x64, gasprice(), gasprice()))
+                    mstore(0x40, m) // Restore the free memory pointer.
+                }
+            }
+            result = _erc1271IsValidSignatureNowCalldata(hash, signature, owner);
```
duncancmt, line 613: ditto previous comment

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,644 @@
...
+            if iszero(call(gas(), proxy, 0x00, 0x10, 0x24, 0x00, 0x00)) {
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
+        }
+    }
+
+    function setPendingOwner(address owner) external onlyFactory {
+        _setPendingOwner(owner);
+    }
```
duncancmt, line 261: let's replace this (and the above selector trick) with `setOwner` rather than `setPendingOwner` because I think that this is introducing pointless additional friction and complexity

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,644 @@
...
+                return IERC1271.isValidSignature.selector;
+            return 0xffffffff;
+        }
+
+        _verifyRoot(MerkleProofLib.getRoot(proof, hash), owner);
+        return IERC1271.isValidSignature.selector;
+    }
+
+    function eip712Domain()
+        public
```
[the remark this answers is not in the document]
e1Ru1o: erc7739 is using it

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,644 @@
...
+            }
+            mstore(0x40, m) // Restore the free memory pointer.
+        }
+        if (t == uint256(0)) hash = _hashTypedData(hash); // `PersonalSign` workflow.
+        result = _erc1271IsValidSignatureNowCalldata(hash, signature, owner);
+    }
+
+    /// @dev Performs the signature validation without nested EIP-712 to allow for easy sign ins.
+    /// This function must always return false or revert if called on-chain.
+    function _erc1271IsValidSignatureViaRPC(bytes32 hash, bytes calldata signature, address owner)
```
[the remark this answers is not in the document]
e1Ru1o: Don't think so, added it to discuss with you. Removing it

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,644 @@
...
+    function _erc1271UnwrapSignature(bytes calldata signature)
+        internal
+        view
+        virtual
+        returns (bytes calldata result)
+    {
+        result = signature;
+        /// @solidity memory-safe-assembly
+        assembly {
+            // Unwraps the ERC6492 wrapper if it exists.
```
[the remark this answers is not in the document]
duncancmt: to be clear, I'm not saying "ditch ERC6492". I'm saying "let's carefully consider whether ERC6492 support is good for either ecosystem compatibility, inspectability, or if it opens up new flows"

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,644 @@
...
+    /// @dev [MODIFIED]Returns whether the `hash` and `signature` are valid.
+    /// Override if you need non-ECDSA logic.
+    function _erc1271IsValidSignatureNowCalldata(bytes32 hash, bytes calldata signature, address owner)
+        internal
+        view
+        virtual
+        returns (bool)
+    {
+        return SignatureCheckerLib.isValidSignatureNowCalldata(owner, hash, signature);
+    }
```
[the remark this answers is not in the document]
e1Ru1o: Assuming for now we don't need

src/CrossChainReceiverFactory.sol
```diff
@@ -156,9 +155,21 @@ contract CrossChainReceiverFactory is IERC1271, MultiCallContext, TwoStepOwnable
...
+            // See: https://eips.ethereum.org/EIPS/eip-7739.
+            // If `hash` is `0x7739...7739`, returns `bytes4(0x77390001)`.
+            // The returned number MAY be increased in future ERC7739 versions.
+            unchecked {
+                if (signature.length == uint256(0)) {
+                    // Forces the compiler to optimize for smaller bytecode size.
+                    if (uint256(hash) == ~signature.length / 0xffff * 0x7739) return 0x77390001;
+                }
+            }
+            if (_erc1271IsValidSignature(hash, signature, owner())) {
```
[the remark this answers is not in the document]
duncancmt: also what happens if `owner()` is `address(0)`? we should prohibit the 7739 path if that is the case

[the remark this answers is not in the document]
e1Ru1o: if it is address(0) it is going to fail because it will not be posible to provide a signature. 

Let me try to change the flow to 
1. Check merkle
2. Check owner
3. Check 7739

src/CrossChainReceiverFactory.sol
```diff
@@ -331,314 +335,86 @@ contract CrossChainReceiverFactory is IERC1271, MultiCallContext, TwoStepOwnable
...
-                calldatacopy(p, add(o, 0x40), c) // Copy `contentsName`, optimistically.
-                mstore(add(p, c), 40) // Store a '(' after the end.
-                if iszero(eq(byte(0, mload(sub(add(p, c), 1))), 41)) {
-                    let e := 0 // Length of `contentsName` in explicit mode.
-                    for { let q := sub(add(p, c), 1) } 1 {} {
-                        e := add(e, 1) // Scan backwards until we encounter a ')'.
-                        if iszero(gt(lt(e, c), eq(byte(0, mload(sub(q, e))), 41))) { break }
+                let p := add(0x0e, m) // Advance 14 bytes to skip "TypedDataSign(".
+                calldatacopy(p, add(0x40, o), c) // Copy `contentsName`, optimistically.
+                mstore(add(p, c), 0x28) // Store a '(' after the end.
```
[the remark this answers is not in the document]
e1Ru1o: It should changing, it was weird to me in a debugging that it was being stored like that but didn't change it 🤦‍♂️

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+                // the ecosystem. This also means that the sort order of the hash and the chainid is
+                // backwards from what `MerkleProofLib` produces, again protecting us against extension
+                // attacks.
+                mstore(0x00, hash)
+                mstore(0x20, chainid())
+                leaf := keccak256(0x00, 0x40)
+
+                // Following assembly blocks are equivalent to:
+                //     (owner, proof) = abi.decode(signature, (address, bytes32[]));
+                // except we omit all the range and overflow checking.
```
duncancmt, line 155: This assembly block doesn't match the code below. Also this needs an explanation around the assumptions that make `validOwner` a good semantic fit. In other words, under what circumstances might the encoding for a merkle signature collide with a 7739 signature?

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+            // If `hash` is `0x7739...7739`, returns `bytes4(0x77390001)`.
+            // The returned number MAY be increased in future ERC7739 versions.
+            unchecked {
+                if (signature.length == uint256(0)) {
+                    // Forces the compiler to optimize for smaller bytecode size.
+                    if (uint256(hash) == ~signature.length / 0xffff * 0x7739) return 0x77390001;
+                }
+            }
+            address owner_ = owner();
+            if ((owner_ != address(0)) && _erc1271IsValidSignature(hash, signature, owner_)) {
```
duncancmt, line 185: Is the check `owner_ != address(0)` required?

duncancmt (commented):
I think I'm getting closer to a complete understanding of this system. It's surprisingly complex.

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+        assembly ("memory-safe") {
+            let ptr := mload(0x40) // Grab the free memory pointer.
+            // Skip 2 words for the `typedDataSignTypehash` and `contents` struct hash.
+            mstore(add(0x40, ptr), nameHash)
+            mstore(add(0x60, ptr), chainid())
+            mstore(add(0x80, ptr), address())
+            
+            // `c` is `contentsDescription.length`, which is stored in the last 2 bytes of the signature.
+            let c := shr(0xf0, calldataload(add(signature.offset, sub(signature.length, 0x02))))
+            for {} 1 {} {
```
duncancmt: I don't understand the heuristic applied to this scoping. Why are some things inside the `for` and others outside?

src/CrossChainReceiverFactory.sol
e1Ru1o: Tried to use "Following assembly block**s**" (with the s) and another comment in the next block saying "continuation" 😅

e1Ru1o: It is not a good a good semantic, basically if it is not an address we skip that section, if it is a valid address we try and fallback to erc7739, there should not be collisions

src/CrossChainReceiverFactory.sol
e1Ru1o: kind of a way to skip the whole validation flow but we can remove it if you don't see it is useful

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    function _erc1271IsValidSignature(bytes32 hash, bytes calldata signature, address owner)
+        internal
+        view
+        virtual
+        returns (bool result)
+    {
+        bytes32 nameHash = _NAMEHASH;
+        assembly ("memory-safe") {
```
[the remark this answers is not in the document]
e1Ru1o: probably yes, let me add it

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+                mstore(p, " contents,string name,uint256 ch") // Store the rest of the encoding.
+                mstore(add(0x20, p), "ainId,address verifyingContract)")
+                p := add(0x40, p)
+                calldatacopy(p, add(0x40, o), c) // Copy `contentsType`.
+                // Fill in the missing fields of the `TypedDataSign`.
+                calldatacopy(ptr, o, 0x40) // Copy the `contents` struct hash to `add(ptr, 0x20)`.
+                mstore(ptr, keccak256(m, sub(add(p, c), m))) // Store `typedDataSignTypehash`.
+                // The "\x19\x01" prefix is already at 0x00.
+                // `APP_DOMAIN_SEPARATOR` is already at 0x20.
+                nameHash := keccak256(ptr, 0xa0)
```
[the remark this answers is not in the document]
e1Ru1o: umm leftover from a debugging round, good catch

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+                signature.length := sub(signature.length, l) // Truncate the signature.
+                
+                // Signature is expected to be 64 bytes
+                if xor(0x40, signature.length) { break }
+                let vs := calldataload(add(0x20, signature.offset))
+
+                mstore(0x00, hash)
+                mstore(0x20, add(0x1b, shr(0xff, vs))) // `v`.
+                mstore(0x40, calldataload(signature.offset)) // `r`.
+                mstore(0x60, shr(0x01, shl(0x01, vs))) // `s`.
```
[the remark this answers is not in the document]
e1Ru1o: Can do, I saw this and left it like that wondering if it was better besides it was not what we have everywhere else

src/CrossChainReceiverFactory.sol
e1Ru1o: now that you mention it, I think the fork might not even be needed. Checking it out tomorrow

duncancmt: this trick with the `for` is typically used because it coaxes solc into generating more efficient code for the branching. it's preferred anytime there's control flow that's more complex than a simple `if` without `else`. it's worth testing, though, now that we've significantly altered the original code

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+            uint256[] memory
+        )
+    {
+        fields = bytes1(0x0d);
+        name_ = name;
+        chainId = block.chainid;
+        verifyingContract = address(this);
+    }
+
+    function deploy(bytes32 root, address owner, bool setOwner) external noDelegateCall returns (address proxy) {
```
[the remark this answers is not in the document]
e1Ru1o: changed to `action`

src/CrossChainReceiverFactory.sol
e1Ru1o: yep yep, looking at it again it is still needed as there are still some relevant negative branches inside the `for`. Theoretically having some code inside or outside the for should not change the output that much 🤔. In this for specifically what I would expect are some `JUMPI` after each `break` but what is inside or outside of the for should not affect the bytecode 🤔

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+            }
+            address pendingOwner_ = pendingOwner();
+            if ((pendingOwner_ == address(0)).or(beneficiary != pendingOwner_)) {
+                _permissionDenied();
+            }
+        }
+        selfdestruct(beneficiary);
+    }
+
+    function _erc1271IsValidSignature(bytes32 hash, bytes calldata signature, address owner)
```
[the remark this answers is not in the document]
e1Ru1o: Changed to `_verifyTypedData` and `_verifyDeploymentData`

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,406 @@
...
+                let o := add(signature.offset, sub(signature.length, l)) // Offset of appended data.
+                mstore(0x00, 0x1901) // Store the "\x19\x01" prefix.
+                calldatacopy(0x20, o, 0x40) // Copy the `APP_DOMAIN_SEPARATOR` and `contents` struct hash.
+                // Dismiss the signature if the reconstructed hash doesn't match,
+                // or if the appended data is invalid, i.e.
+                // `appendedData.length > signature.length || contentsDescription.length == 0`.
+                if or(xor(keccak256(0x1e, 0x42), hash), or(lt(signature.length, l), iszero(c))) {
+                    break
+                }
+                // Else, use the `TypedDataSign` workflow.
```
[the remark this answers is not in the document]
e1Ru1o: "Workflow" is misleading indeed as there is only one, changing it

src/CrossChainReceiverFactory.sol
e1Ru1o: Ended up removing the `for` anyway

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    // Modified from Solady (https://github.com/Vectorized/solady/blob/c4d32c3e6e89da0321fda127ff024eecd5b57bc6/src/accounts/ERC1271.sol#L120-L287)
+    function _verifyTypedDataSignature(bytes32 hash, bytes calldata signature, address owner)
+        internal
+        view
+        virtual
+        returns (bool result)
+    {
+        bytes32 nameHash = _NAMEHASH;
```
duncancmt: we might get more optimized code by setting `_NAMEHASH` to a literal (check that it matches in the constructor) and referencing it directly

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,400 @@
...
+                        if iszero(gt(lt(e, c), eq(byte(0x00, mload(sub(q, e))), 0x29))) { break }
+                    }
+                    c := sub(c, e) // Truncate `contentsDescription` to `contentsType`.
+                    calldatacopy(p, add(add(0x40, o), c), e) // Copy `contentsName`.
+                    mstore8(add(p, e), 0x28) // Store a '(' exactly right after the end.
+                }
+                // `d & 1 == 1` means that `contentsName` is invalid.
+                let d := shr(byte(0, mload(p)), 0x7fffffe000000000000010000000000) // Starts with `[a-z(]`.
+                // Advance `p` until we encounter '('.
+                for {} iszero(eq(byte(0x00, mload(p)), 0x28)) { p := add(0x01, p) } {
```
duncancmt: I wonder if solc is smart enough to optimize this or if the following would be more efficient
```suggestion
                for {} xor(0x28, shr(0xf8, mload(p))) { p := add(0x01, p) } {
```

duncancmt: it might also be a good idea to substitute this for the equivalent expression in the body of the loop below

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,400 @@
...
+                // backwards from what `MerkleProofLib` produces, again protecting us against extension
+                // attacks.
+                mstore(0x00, hash)
+                mstore(0x20, chainid())
+                leaf := keccak256(0x00, 0x40)
+
+                // In merkle proof validation flow, signature is formed by (owner, proof)
+                // Following assembly decodes the owner and checks it is a valid address
+                owner := calldataload(signature.offset)
+                validOwner := iszero(shr(0xa0, owner))
```
duncancmt: This comment is just here because the old comment got dropped because it's outdated. This comment needs an explanation about why the two different signature schemes can be disambiguated with high confidence and the number of bits the attacker would need to control in order for there to be confusion

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,400 @@
...
+            // 1. the reconstructed hash doesn't match,
+            // 2. the appended data is invalid, i.e. 
+            //    (`appendedData.length > signature.length || contentsDescription.length == 0`.)
+            // 3. the signature is not 64 bytes long
+            if iszero(or(xor(keccak256(0x1e, 0x42), hash), or(xor(add(0x40, l), signature.length), iszero(c)))) {
+                // Generate the `TypedDataSign` struct.
+                // `TypedDataSign({ContentsName} contents,string name,...){ContentsType}`.
+                // and check it was signed by the owner
+                let m := add(0xa0, ptr)
+                mstore(m, "TypedDataSign(") // Store the start of `TypedDataSign`'s type encoding.
```
duncancmt, line 508: idle thought, feel free to ignore. the literal `"TypedDataSign("` is right-padded. can we substitute this literal with one that is left-padded without sacrificing runtime gas in order to improve contract size?

src/CrossChainReceiverFactory.sol
duncancmt: can you double check that this actually improves gas usage? the `for` trick tends to be the most optimal

e1Ru1o: I can check but notice that now there is a single if without else, that was exactly the case you mentioned
> it's preferred anytime there's control flow that's more complex than a simple if without else

If the `for` is better solc really does crazy stuff 😅

duncancmt: > solc really does crazy stuff 😅

unfortunately, this is the case :crying_cat_face: it's always best to check by running the test and noting the gas

src/CrossChainReceiverFactory.sol
e1Ru1o: sorry I don't see the part related to the body of the loop 🤔

src/CrossChainReceiverFactory.sol
e1Ru1o: Doesn't seem to change that much
```
| Contract                            | Runtime Size (B) | Initcode Size (B) | Runtime Margin (B) | Initcode Margin (B) |
| CrossChainReceiverFactory (current) | 3,730            | 4,565             | 20,846             | 44,587              |
| CrossChainReceiverFactory (new)     | 3,730            | 4,567             | 20,846             | 44,585              |
```

src/CrossChainReceiverFactory.sol
duncancmt: I was referring to the use of the `byte(0x00, mload(p))`, but now that I reread it, I think it's optimal as-is

src/CrossChainReceiverFactory.sol
duncancmt: how does it change the runtime gas?

src/CrossChainReceiverFactory.sol
e1Ru1o: yep I didn't find a way to optimize it

duncancmt: did my suggestion for this line improve gas consumption? or did `solc` outsmart me?

src/CrossChainReceiverFactory.sol
```diff
@@ -17,7 +17,7 @@ contract CrossChainReceiverFactory is IERC1271, IERC5267, MultiCallContext, TwoS
     CrossChainReceiverFactory private immutable _cachedThis;
     bytes32 private immutable _proxyInitHash;
     string public constant name = "ZeroExCrossChainReceiver";
-    bytes32 private constant _NAMEHASH = 0x819c7f86c24229cd5fed5a41696eb0cd8b3f84cc632df73cfd985e8b100980e8;
+    uint256 private constant _NAMEHASH = 0x819c7f86c24229cd5fed5a41696eb0cd8b3f84cc632df73cfd985e8b100980e8;
```
duncancmt: I think this works both ways? It shouldn't matter whether it's declared as `uint256` or `bytes32` if all you're going to do is reference it in Yul. Yul doesn't have types

src/CrossChainReceiverFactory.sol
```diff
@@ -349,7 +348,7 @@ contract CrossChainReceiverFactory is IERC1271, IERC5267, MultiCallContext, TwoS
                 // `TypedDataSign({ContentsName} contents,string name,...){ContentsType}`.
                 // and check it was signed by the owner
                 let m := add(0xa0, ptr)
-                mstore(m, "TypedDataSign(") // Store the start of `TypedDataSign`'s type encoding.
+                mstore(m, 0x5479706564446174615369676e28000000000000000000000000000000000000) // Store the start of `TypedDataSign`'s type encoding.
```
duncancmt: I don't understand the point of this change. Isn't this identical?

src/CrossChainReceiverFactory.sol
e1Ru1o: 6964 with the `if` vs 7376 with the `for`

Keeping it as it is

duncancmt: That's a huge difference! I am very suspicious

src/CrossChainReceiverFactory.sol
```diff
@@ -79,9 +75,6 @@ contract CrossChainReceiverFactory is IERC1271, IERC5267, MultiCallContext, TwoS
     IWrappedNative private immutable _WNATIVE =
         IWrappedNative(payable(address(uint160(uint256(bytes32(_WNATIVE_STORAGE.code))))));
 
-    error DeploymentFailed();
-    error ApproveFailed();
-
```
duncancmt: They're not used in a `revert` statement, but they're still part of the ABI. It's just that they're always thrown from assembly

e1Ru1o: Noted that we keep them as part of the ABI, sorry for removing them

src/CrossChainReceiverFactory.sol
e1Ru1o: you were right on being suspicious, just double checked, chaining it to the `for`. It saves a bit of gas, kept as much code outside of the `for` as posible because it was gives the best gas results.

.github/workflows/test.yml
e1Ru1o, line 1: Maybe remove the `-vvvv` in the action

src/CrossChainReceiverFactory.sol
```diff
@@ -0,0 +1,471 @@
...
+                    // encoded anonymous struct `(owner, proof)`. It omits range and overflow
+                    // checking.
+                    proof.offset := add(signature.offset, calldataload(add(0x20, signature.offset)))
+                    proof.length := calldataload(proof.offset)
+                    proof.offset := add(0x20, proof.offset)
+                }
+
+                return _verifyDeploymentRootHash(MerkleProofLib.getRoot(proof, hash), originalOwner)
+                    ? IERC1271.isValidSignature.selector
+                    : bytes4(0xffffffff);
```
e1Ru1o: NIce that we don't need to fallback to ERC7739 anymore

e1Ru1o:
Looks good to proceed to the audit

e1Ru1o:
iirc we discussed merging into a separated branch. I am going to create `bridge` and point the future bridge related PRs to that one

src/Settler.sol
```diff
@@ -115,18 +115,24 @@ abstract contract Settler is ISettlerTakerSubmitted, Permit2PaymentTakerSubmitte
...
         if (actions.length != 0) {
-            (uint256 action, bytes calldata data) = actions.decodeCall(0);
-            if (!_dispatchVIP(action, data)) {
-                if (!_dispatch(0, action, data)) {
-                    revertActionInvalid(0, action, data);
-                }
+            uint256 it;
+            assembly ("memory-safe") {
+                it := actions.offset
             }
```
duncancmt: Can you abstract this pattern into a library to enforce the abstraction barrier? Something like this https://github.com/0xProject/0x-settler/blob/613dc703d169d55360704c907c64d5063ecfe64c/src/core/EulerSwap.sol#L490-L526

e1Ru1o: might be skill issue on my side but I am not making it to add the abstraction without increasing code size 😅

duncancmt: Hmm... It should be a zero-cost abstraction. Can you push what you've got up to another branch so that I can take a look?

e1Ru1o: Agree that it should be zero-cost. Here it is https://github.com/0xProject/0x-settler/commit/65d1441ae18725c5768e5977d3d4de953a3e9798

duncancmt: I tried a bunch of different ways. You're totally right, for some reason solc isn't making this abstraction zero-cost :(

src/core/MakerPSM.sol
```diff
@@ -96,31 +98,31 @@ abstract contract MakerPSM is SettlerAbstract {
         internal
         returns (uint256 buyAmount)
     {
+        (IERC20 sellToken, IERC20 buyToken) = buyGem.maybeSwap(USDC, DAI);
```
duncancmt, line 115: One thing that we need to support (ultimately) is the USDS PSM (`0xA188EEC8F81263234dA3622A406892F3D630f98c`). How much work/size would it be to refactor this to make it generic between the existing "Lite" PSM and the new Sky PSM?

src/Settler.sol
e1Ru1o: super sad

src/core/MakerPSM.sol
e1Ru1o: What do you expect here? Seems like we need a new param to differentiate between lite and Sky, it should not be that much. Options as I see them:
1. Provide the address and we check it is not a restricted target. This way removes the constants for Lite and Sky
2. Provide a bool indicating which of the two to use.
What do you prefer? I bet both work, I can try them out and report back with the numbers.

Do you have any other option in mind?

PR #340, opened 2025-05-01
Golf BridgeFactory and make `cleanup` work on Cancun

duncancmt:


src/BridgeFactory.sol
```diff
@@ -36,70 +48,66 @@ contract BridgeFactory is IERC1271, MultiCallContext, TwoStepOwnable {
...
-        return _isValidSignature(_hash, _signature);
-    }
-
-    function _isValidSignature(bytes32 _leaf, bytes calldata _proof) private view returns (bytes4) {
         bytes32[] calldata proof;
         assembly ("memory-safe") {
-            // _signature is just going to be the proof, then we can read it as so
-            proof.offset := add(0x20, _proof.offset)
+            // signature is just the proof, then we can read it as so
+            proof.offset := add(signature.offset, calldataload(signature.offset))
```
e1Ru1o, line 92: Can we not assume here it is going to be `0x20`? 🤔

src/BridgeFactory.sol
```diff
@@ -36,70 +48,66 @@ contract BridgeFactory is IERC1271, MultiCallContext, TwoStepOwnable {
...
-
-            if shl(0x60, xor(computedAddress, address())) {
-                mstore(0x00, 0x00)
-                return(0x00, 0x20)
-            }
-            // Return ERC1271 magic value (isValidSignature selector)
-            mstore(0x00, 0x1626ba7e00000000000000000000000000000000000000000000000000000000)
-            return(0x00, 0x20)
-        }
+        _verifyRoot(MerkleProofLib.getRoot(proof, hash), pendingOwner());
```
e1Ru1o, line 97: so if the ownership gets claimed, everything gets invalidated from that moment on. 👍

src/BridgeFactory.sol
duncancmt: that's true if the abiencoding is strict, but since anybody may call `isValidSignature`, I don't think we can safely assume that

src/BridgeFactory.sol
```diff
@@ -36,70 +48,66 @@ contract BridgeFactory is IERC1271, MultiCallContext, TwoStepOwnable {
...
-            mstore(0x60, 0x00)
-
-            if xor(owner, mul(mload(0x20), eq(returndatasize(), 0x20))) {
-                mstore(0x00, 0x8baa579f) // selector for `InvalidSignature()`
-                revert(0x1c, 0x04)
-            }
-        
-            mstore(0x20, chainid())
-            salt := keccak256(0x00, 0x40)
+            mstore(0x14, owner)
```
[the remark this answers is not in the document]
duncancmt: yep. good catch

src/BridgeFactory.sol
duncancmt: yeah, I think that's a useful feature. of course, there's the weird edge case where if you `transferOwnership` back to the original owner, all of a sudden some signature again become valid. I think that I'm not gonna fix that, though

e1Ru1o (approved):

src/BridgeFactory.sol
```diff
@@ -111,29 +120,57 @@ contract BridgeFactory is IERC1271, MultiCallContext, TwoStepOwnable {
...
+            if (owner_ != address(0)) {
+                assembly ("memory-safe") {
+                    mstore(0x00, 0x1e092104) // selector for `PermissionDenied()`
+                    revert(0x1c, 0x04)
+                }
+            }
+            address pendingOwner_ = pendingOwner();
+            if (pendingOwner_ == address(0)) {
+                selfdestruct(beneficiary);
+            }
```
duncancmt, line 171: @e1Ru1o what do you think of this clause? I'm kinda torn on it. This effectively means that on pre-Cancun chains, calling `renounceOwnership` really just resets the owner to the initial owner. I have less of a feeling about the fact that this means that the ETH in a renounced proxy can be taken by anybody, because that ETH was already basically lost.

e1Ru1o: I see you changed it now. I would let only the owner call cleanup, if the `pendingOwner` wants the cleanup to be done he needs to first `acceptOwnership`. This is in favor of keeping the logic clean/simple.

I see as well the logic for calling cleanup from `deploy`, not a fan at the moment, it seems to me it is for rescuing ETH that was sent to proxy after it was already cleaned up, is that so? if so, it is a very intricate case, I would not expect assets being sent there anymore so if it happens I will let the user do the rescue. You can have a separated `rescue` function for that if you want that does the `deploy` + `cleanup`, but even without it a user can do it.

e1Ru1o (approved):
Changes look good

PR #341, opened 2025-05-06
Single Signature Hack Tests

e1Ru1o:


test/unit/SingleSignatureDirtyHack.t.sol
```diff
@@ -0,0 +1,386 @@
...
+        address operator,
+        bytes memory extraData,
+        function(uint256, uint256, uint256, address, bytes memory, bytes memory) view returns (bytes32) getPayload,
+        function(address, SingleSignatureDirtyHack.TransferParams memory, uint256, uint256, PackedSignature memory, bytes memory) internal
+            transfer
+    ) internal {
+        signerPk = bound(signerPk, 2, 115792089237316195423570985008687907852837564279074904382605163141518161494336);
+        gasLimit = bound(gasLimit, 0, 30_000_000);
+        deadline = bound(deadline, block.timestamp, type(uint256).max - 1);
+        nonce = bound(nonce, 1, type(uint64).max - 3);
```
duncancmt, line 287: why is `nonce == 0` outside the scope of the test? and isn't the maximum nonce `type(uin64).max - 1`?

test/unit/SingleSignatureDirtyHack.t.sol
```diff
@@ -0,0 +1,386 @@
...
+        });
+
+        vm.prank(transferParams.from);
+        token.approve(address(harness), amount);
+        deal(address(token), address(transferParams.from), amount);
+
+        transfer(operator, transferParams, gasPrice, gasLimit, sig, extraData);
+
+        // can use incremental nonce
+        nonce++;
```
duncancmt, line 312: why is `nonce == 1` outside the scope of the test?

duncancmt (commented):
very nice. very compact!

test/unit/SingleSignatureDirtyHack.t.sol
e1Ru1o: 0 is outside because it is not allowed in the current implementation
image

As for the maximum it can be `type(uin64).max - 2` as I increment the nonce to times in the tests. Changing that.

test/unit/SingleSignatureDirtyHack.t.sol
e1Ru1o: this test is for proving that after a valid txn doing another with a greater nonce works. Notice that before this line a transfer was already accepted so only nonces from `nonce + 1` are valid

test/unit/SingleSignatureDirtyHack.t.sol
```diff
@@ -0,0 +1,386 @@
...
+        uint256 requestedAmount,
+        bytes32 salt,
+        address operator,
+        bytes memory extraData,
+        function(uint256, uint256, uint256, address, bytes memory, bytes memory) view returns (bytes32) getPayload,
+        function(address, SingleSignatureDirtyHack.TransferParams memory, uint256, uint256, PackedSignature memory, bytes memory) internal
+            transfer
+    ) internal {
+        signerPk = bound(signerPk, 2, 115792089237316195423570985008687907852837564279074904382605163141518161494336);
+        gasLimit = bound(gasLimit, 0, 30_000_000);
```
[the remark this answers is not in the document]
e1Ru1o: I would say both checks minimum and maximum can be skipped, providing a proper value for the gasLimit is responsibility of whoever craft the message to be signed, so that is some gas that you can save onchain. Having the checks doesn't ensure correctness.

test/unit/SingleSignatureDirtyHack.t.sol
```diff
@@ -0,0 +1,386 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {Test} from "@forge-std/Test.sol";
+import {IERC20} from "@forge-std/mocks/MockERC20.sol";
+import {LibRLP} from "@solady/utils/LibRLP.sol";
```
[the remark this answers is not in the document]
e1Ru1o: Let me check

test/unit/SingleSignatureDirtyHack.t.sol
duncancmt: oh! nice catch. :thinking: I wonder how we can solve this. maybe we need to store the next nonce, rather than the current nonce, so that the comparison can be `<` instead of `<=`. we definitely want a nonce of zero to be allowed

test/unit/SingleSignatureDirtyHack.t.sol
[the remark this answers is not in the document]
duncancmt: this feels like it's becoming a question for an auditor. we want to only accept the same signatures that Ethereum itself would accept. is there an attack that can be performed if we were to accept a signature that's _almost_ a valid Ethereum transaction?

test/unit/SingleSignatureDirtyHack.t.sol
e1Ru1o: yep just storing the `incomingNonce + 1` do the work, let me change it

e1Ru1o: Also kept `type(uint64).max - 3` as max because I increment twice and `type(uint64).max` is not allowed

test/unit/SingleSignatureDirtyHack.t.sol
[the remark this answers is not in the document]
e1Ru1o: There will be a lot of checks to make that happen. For example, for EIP1559, we will need to check `max_priority_fee_per_gas < max_fee_per_gas`, which in our code will be `gasPriorityPrice < gasPrice`.
There will probably be more checks like that we need to add if we want to only allow correct txns

test/unit/SingleSignatureDirtyHack.t.sol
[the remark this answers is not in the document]
e1Ru1o: Used ethers to generate the txns.

src/SingleSignatureDirtyHack.sol
```diff
@@ -120,8 +120,8 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
     // allowance.
     function _consumeNonce(address owner, uint256 incomingNonce) private {
         uint256 currentNonce = nonces[owner];
-        nonces[owner] = incomingNonce;
-        if (incomingNonce <= currentNonce) {
+        nonces[owner] = incomingNonce + 1;
```
duncancmt, line 123: can this be unchecked? do we need different range bounds on the nonce to avoid overflowing the valid nonce range, smaller than the uint256 range?

duncancmt (approved):

duncancmt:
I can imagine some clever techniques involving checking the receipt of a transaction against the block header hash, but I don't want to do that because verkle will definitely break that. so there's necessarily going to be some gaps in the implementation

src/SingleSignatureDirtyHack.sol
e1Ru1o: I thought that was intentional, I would say the standard is the nonces mapping just to admit any nonce in any order. Fine with that?

test/unit/SingleSignatureDirtyHack.t.sol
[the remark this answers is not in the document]
e1Ru1o: After reading your las review comment
> I can imagine some clever techniques involving checking the receipt of a transaction against the block header hash, but I don't want to do that because verkle will definitely break that. so there's necessarily going to be some gaps in the implementation

I got what was worrying you. And indeed, what we would need to do is check the receipt/inclusion of the txn instead of its correctness

src/SingleSignatureDirtyHack.sol
duncancmt: > that was intentional
that it's possible for the user to grief themself? I wouldn't say intentional, but I also don't see a good solution.

do we need to restrict the range of acceptable nonces (i.e., <= type(uint64).max - 1)? can we use `unchecked` here?

e1Ru1o: On one side, we can safely decrease to uint64, but we don't need to. We can do unchecked and revert if it becomes 0 (overflow) in the current implementation.

As for the griefing I see 2 options:
1. You assume it is incremental and every transaction is going to be consumed right after sending the approve one so signing nonce X invalidates all previous ones (current implementations)
2. You assume the `transferFrom` execution in our contract can be done unordered, then allow unordered nonces.

Number 2 can be the user submitting multiple approve trick transactions, and then "consuming" them with transferFrom in any order.

Guess this will also be improved (or removed) once we check receipts/inclusion.

e1Ru1o: Actually it doesn't makes sense to consume unordered

src/SingleSignatureDirtyHack.sol
```diff
@@ -120,9 +120,16 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+        unchecked {
+            uint256 nextNonce = incomingNonce + 1;
+            // for incomingNonce being MAX_UINT256, nextNonce will underflow and become 0
+            // but currentNonce + 1 will not overflow as it starts at 0 and only changes after
+            // accepting an incoming nonce, then, next condition will be true as
+            // currentNonce + 1 will be positive. MAX_UINT256 is not allowed.
+            if (nextNonce < currentNonce + 1) {
+                revert NonceReplay(currentNonce, incomingNonce);
+            }
+            nonces[owner] = nextNonce;
```
duncancmt, line 130: put this line immediately after computing `nextNonce` to avoid duplicate `keccak256`

duncancmt (approved):

PR #343, opened 2025-05-07
Add permit trick

e1Ru1o:


src/SingleSignatureDirtyHack.sol
```diff
@@ -198,15 +223,87 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
         uint256 gasPrice,
         uint256 gasLimit,
         PackedSignature calldata sig
     ) external preFlightChecklist(transferParams) returns (bool) {
         bytes32 signingHash = _hashStruct(typeSuffix, transferParams);
-        bytes memory data = _encodeData(transferParams.amount, signingHash);
+        bytes memory data = _encodePermitData(transferParams.from, transferParams.amount, signingHash, sig);
         address signer = TransactionEncoder.recoverSigner(
             transferParams.nonce, gasPrice, gasLimit, payable(address(transferParams.token)), 0 wei, data, sig
         );
```
duncancmt, line 309: I don't think this is what you want. We don't want to recover the signature of the RLP encoded transaction. We want to recover the signer of the EIP712 `Permit` object (including token domain separator)

e1Ru1o: Updated

src/SingleSignatureDirtyHack.sol
```diff
@@ -187,6 +195,63 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            // clear clobbered memory
+            mstore(0x60, 0x00)
+            mstore(0x80, ptr2)
+            mstore(0xa0, ptr3)
+
+            let structHash := keccak256(0x00, 0x80)
+
+            mstore(0x00, 0x3644e515) // selector for `DOMAIN_SEPARATOR()`
+            if iszero(staticcall(gas(), token, 0x1c, 0x04, 0x20, 0x20)) {
+                mstore(0x40, ptr1)
```
duncancmt, line 236: what's the point of this line?

src/SingleSignatureDirtyHack.sol
```diff
@@ -187,6 +195,63 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            let ptr1 := mload(0x40)
+            let ptr2 := mload(0x80)
+            let ptr3 := mload(0xa0)
+
+            mstore(0x00, permitTypeHash)
+            mstore(0x20, from)
+            mstore(0x40, address())
+            mstore(0x60, sellAmount)
+            mstore(0x80, nonce)
+            mstore(0xa0, signingHash)
```
duncancmt, line 225: you must never clobber the memory in the range `0x80` to `mload(0x40)`. that is memory reserved by the compiler for spilling variables from the stack

src/SingleSignatureDirtyHack.sol
```diff
@@ -166,13 +167,20 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
         _;
     }
 
+    function _checkNonce(address token, address owner, uint256 previousNonce) private view {
+        uint256 currentNonce = EIP2612(token).nonces(owner);
```
duncancmt, line 171: consider rewriting this line in assembly for efficiency/size

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +334,27 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            transferParams.token, transferParams.from, transferParams.nonce, transferParams.amount, structHash
+        );
+        (bytes32 r, bytes32 vs) = (sig.r, sig.vs);
+        uint8 v;
+        unchecked {
+            v = uint8(uint256(vs) >> 255) + 27;
+        }
+        bytes32 s;
+        s = vs & 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
+        _checkSigner(transferParams.from, ecrecover(signingHash, v, r, s));
```
duncancmt, line 355: consider reimplementing this `ecrecover` in assembly for efficiency/size

src/interfaces/EIP2612.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+bytes32 constant PERMIT_TYPEHASH =
+    keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
+
+interface EIP2612 {
```
duncancmt, line 7:
```suggestion
interface IEIP2612 is IERC20 {
```

duncancmt (changes requested):

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +334,27 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
         transferParams.token.safeTransferFrom(transferParams.from, transferParams.to, transferParams.requestedAmount);
         return true;
     }
+
+    function transferFromPermit(
+        string calldata typeSuffix,
+        TransferParams calldata transferParams,
+        PackedSignature calldata sig
+    ) external preFlightChecklist(transferParams) returns (bool) {
```
duncancmt, line 342: `consumeNonce` is inapplicable here, so we can't use `preFlightChecklist`

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +334,27 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+        _checkNonce(address(transferParams.token), transferParams.from, transferParams.nonce);
+        bytes32 structHash = _hashStruct(typeSuffix, transferParams);
+        bytes32 signingHash = _encodePermitHash(
+            transferParams.token, transferParams.from, transferParams.nonce, transferParams.amount, structHash
+        );
```
duncancmt, line 347: you might want to consider actually making the `permit` call as in this contract https://github.com/0xProject/permitandcall/blob/0977908b58cd82ade39527389399221248821572/src/lib/SafePermit.sol#L91-L127

but if you do, be sure not to bubble any revert, because if you do it creates a griefing opportunity.

duncancmt (commented):
some comments from my earlier review didn't post

src/SingleSignatureDirtyHack.sol
e1Ru1o: right, didn't notice 🤦‍♂️

src/SingleSignatureDirtyHack.sol
e1Ru1o: Saw it earlier but, doing exactly what is done there breaks the flow described in Notion of doing a multicall with the permit call and the call to BridgeSettler. Given your suggestion we will do the permit call ourselves? Please confirm before I change it. It seems definitely fine to do anyway

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +340,40 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            mstore(0x00, signingHash)
+            mstore(0x20, add(0x1b, shr(0xff, vs))) // v
+            mstore(0x40, calldataload(sig)) // r
+            mstore(0x60, and(vs, 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)) // s
+            
+            if iszero(staticcall(gas(), 0x01, 0x00, 0x80, 0x00, 0x20)) {
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
+            signer := mload(0x00)
```
duncancmt, line 368: the `ecrecover` precompile only fails due to out-of-gas

```suggestion
            pop(staticcall(gas(), 0x01, 0x00, 0x80, 0x00, 0x20))
            signer := mul(mload(0x00), eq(returndatasize(), 0x20))
```

src/SingleSignatureDirtyHack.sol
```diff
@@ -166,13 +165,36 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            mstore(0x14, owner) 
+            mstore(0x00, 0x7ecebe00000000000000000000000000) // selector for `nonces(address)` with `owner` padding
+            if iszero(staticcall(gas(), token, 0x10, 0x24, 0x00, 0x20)) {
+                let ptr := mload(0x40)
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
+            if lt(returndatasize(), 0x20) {
+                mstore(0x00, 0xc1ab6dc1) // selector for `InvalidToken()`
+                revert(0x1c, 0x04)
```
duncancmt, line 179: it's fine to revert with an empty reason here. that mimics the behavior of native Solidity under the same conditions

src/SingleSignatureDirtyHack.sol
```diff
@@ -187,6 +209,55 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+
+            mstore(0x00, 0x3644e515) // selector for `DOMAIN_SEPARATOR()`
+            if iszero(staticcall(gas(), token, 0x1c, 0x04, 0x20, 0x20)) {
+                returndatacopy(ptr, 0x20, returndatasize())
+                revert(ptr, returndatasize())
+            }
+            if lt(returndatasize(), 0x20) {
+                mstore(0x00, 0xc1ab6dc1) // selector for `InvalidToken()`
+                revert(0x1c, 0x04)
+            }
```
duncancmt, line 249: ditto above comment

src/SingleSignatureDirtyHack.sol
```diff
@@ -187,6 +209,55 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+        //             value,
+        //             nonce,
+        //             deadline))
+        // ))
+        bytes32 permitTypeHash = PERMIT_TYPEHASH;
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+
+            mstore(ptr, permitTypeHash)
+            mstore(add(0x20, ptr), from)
```
duncancmt, line 233: `from` has dirty bits

src/SingleSignatureDirtyHack.sol
```diff
@@ -166,13 +165,36 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            if iszero(staticcall(gas(), token, 0x10, 0x24, 0x00, 0x20)) {
+                let ptr := mload(0x40)
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
+            if lt(returndatasize(), 0x20) {
+                mstore(0x00, 0xc1ab6dc1) // selector for `InvalidToken()`
+                revert(0x1c, 0x04)
+            }
+            if iszero(eq(mload(0x00), add(0x01, previousNonce))) {
```
duncancmt, line 181:
```suggestion
            if xor(mload(0x00), add(0x01, previousNonce)) {
```

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +340,40 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+        );
+        address signer;
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            let vs := calldataload(add(0x20, sig))
+
+            mstore(0x00, signingHash)
+            mstore(0x20, add(0x1b, shr(0xff, vs))) // v
+            mstore(0x40, calldataload(sig)) // r
+            mstore(0x60, and(vs, 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)) // s
```
duncancmt, line 362: the compiler is probably smart enough to sort this out, but for consistency
```suggestion
            mstore(0x60, and(0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff, vs)) // s
```

src/SingleSignatureDirtyHack.sol
duncancmt: Good point :thinking: 

I think it's preferable if we do it ourselves because then we can avoid the extra cold account access of going through MultiCall as well as the copying/encoding overhead. What do you think?

duncancmt (approved):

src/SingleSignatureDirtyHack.sol
e1Ru1o: Sounds good to me, doing the change

src/SingleSignatureDirtyHack.sol
e1Ru1o: ok will revert this then

src/SingleSignatureDirtyHack.sol
```diff
@@ -187,6 +186,52 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            mstore(add(0x40, ptr), address())
+            mstore(add(0x60, ptr), sellAmount)
+            mstore(add(0x80, ptr), nonce)
+            mstore(add(0xa0, ptr), signingHash)
+
+            let structHash := keccak256(ptr, 0xc0)
+
+            mstore(0x00, 0x3644e515) // selector for `DOMAIN_SEPARATOR()`
+            if iszero(staticcall(gas(), token, 0x1c, 0x04, 0x20, 0x20)) {
+                returndatacopy(ptr, 0x20, returndatasize())
```
duncancmt, line 220:
```suggestion
                returndatacopy(ptr, 0x00, returndatasize())
```

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +314,90 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            mstore(add(0xd4, ptr), and(vs, 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)) // s
+            mstore(add(0xb4, ptr), calldataload(sig)) // r
+            mstore(add(0x94, ptr), add(0x1b, shr(0xff, vs))) // v
+            mstore(add(0x74, ptr), signingHash)
+            
+            pop(staticcall(gas(), 0x01, add(0x74, ptr), 0x80, 0x00, 0x20))
+            signer := mul(mload(0x00), eq(returndatasize(), 0x20))
+
+            // keep adding parameters to call permit
+            // TODO: check that calling ecrecover doesn't affect the previously added parameters
```
duncancmt, line 347: `mload(0x40)` can never be below `0x80`, so specifying `0x00, 0x20` as the returndata region for `ecrecover` can't clobber any of the values that you put after the free memory pointer

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +314,90 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            mstore(add(0x54, ptr), value)
+            mstore(add(0x34, ptr), address())
+            mstore(add(0x14, ptr), owner)
+            mstore(ptr, 0xd505accf000000000000000000000000) // selector for `permit(address,address,uint256,uint256,uint8,bytes32,bytes32)` with `owner` padding
+
+            let token := and(0xffffffffffffffffffffffffffffffffffffffff, calldataload(add(0x40, transferParams)))
```
duncancmt, line 353: don't clean dirty bits here. dirty bits in calldata should result in a revert with an empty reason. of course, if this checking is performed elsewhere in this context (i.e. the expression `transferParams.token` appears below), then you don't need to clean or revert; we can be confident that the bits are clean

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +314,90 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            if iszero(call(gas(), token, 0x00, add(0x10, ptr), 0xe4, 0x00, 0x00)) {
+                // TODO: Check why not to bubble reverts
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
```
duncancmt, line 358: we can't bubble reverts here. the only check that we can perform is that the call didn't OOG (see `SafeCall.sol` from the reference implementation and a similar pattern in `CheckCall.sol` in this repo). by bubbling the revert here, we open ourselves up to a griefing/frontrunning attack because nonce replays cause a revert in `permit`.

duncancmt (approved):

src/SingleSignatureDirtyHack.sol
```diff
@@ -269,4 +314,90 @@ abstract contract SingleSignatureDirtyHack is IERC5267, AbstractContext {
...
+            if lt(returndatasize(), 0x20) { revert(0x00, 0x00) }
+            if xor(mload(0x00), add(0x01, previousNonce)) {
+                mstore(ptr, 0x1fa72369) // selector for `NonceReplay(uint256,uint256)`
+                mstore(add(0x20, ptr), mload(0x00))
+                mstore(add(0x40, ptr), previousNonce)
+                revert(add(0x1c, ptr), 0x44)
+            }
+
+            // check allowance
+            mstore(add(0x34, ptr), address())
```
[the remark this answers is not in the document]
e1Ru1o: yep got it while testing

src/SingleSignatureDirtyHack.sol
e1Ru1o: dirty bits are killing me haha, why from had dirty bits then in one of your previous comments? it was passed to the function from the calldata

e1Ru1o: ahh it doesn't apply. Got it

PR #354, opened 2025-06-10
Support for Ekubo forwarding extensions

duncancmt:


src/core/DodoV1.sol
```diff
@@ -47,8 +47,9 @@ library FastDodoV1 {
             mstore(add(0x80, ptr), 0x00)
 
             if iszero(call(gas(), dodo, 0x00, add(0x1c, ptr), 0x84, 0x00, 0x20)) {
-                returndatacopy(ptr, 0x00, returndatasize())
-                revert(ptr, returndatasize())
+                let ptr_ := mload(0x40)
+                returndatacopy(ptr_, 0x00, returndatasize())
+                revert(ptr_, returndatasize())
```
e1Ru1o, line 52: why? I see it is now done like that consistently, just want to know

e1Ru1o (approved):

src/core/DodoV1.sol
duncancmt: not having to copy `ptr` from outside the `if` saves ~5 gas in the non-reverting path. it also increases code size a little bit

PR #355, opened 2025-06-10
EulerSwap

duncancmt:
See https://docs.euler.finance/developers/contract-addresses for EVC addresses


duncancmt:
MainnetSettlerIntent needs some size golfing before this can go out, but everything else is in a shippable state

src/core/Permit2Payment.sol
```diff
@@ -29,11 +29,11 @@ import {AllowanceHolderContext} from "../allowanceholder/AllowanceHolderContext.
...
 library TransientStorage {
     // bytes32((uint256(keccak256("operator slot")) - 1) & type(uint128).max)
-    bytes32 private constant _OPERATOR_SLOT = 0x0000000000000000000000000000000007f49fa1cdccd5c65a7d4860ce3abbe9;
+    bytes32 private constant _OPERATOR_SLOT = 0x0000000000000000000000000000000000000000cdccd5c65a7d4860ce3abbe9;
     // bytes32((uint256(keccak256("witness slot")) - 1) & type(uint128).max)
-    bytes32 private constant _WITNESS_SLOT = 0x00000000000000000000000000000000e44a235ac7aebfbc05485e093720deaa;
+    bytes32 private constant _WITNESS_SLOT = 0x0000000000000000000000000000000000000000c7aebfbc05485e093720deaa;
     // bytes32((uint256(keccak256("payer slot")) - 1) & type(uint128).max)
-    bytes32 private constant _PAYER_SLOT = 0x00000000000000000000000000000000c824a45acd1e9517bb0cb8d0d5cde893;
+    bytes32 private constant _PAYER_SLOT = 0x0000000000000000000000000000000000000000cd1e9517bb0cb8d0d5cde893;
```
e1Ru1o: update comments, it is not `type(uint128).max` anymore

src/multicall/MultiCallContext.sol
```diff
@@ -27,37 +29,38 @@ interface IMultiCall {
...
-            sender := and(0xffffffffffffffffffffffffffffffffffffffff, sender)
             // ERC-2771. The trusted forwarder (`_MULTICALL`) has appended the appropriate
             // msg.sender to the msg data
             sender :=
                 xor(
                     sender,
                     mul(
                         xor(shr(0x60, calldataload(add(data.offset, sub(data.length, 0x14)))), sender),
-                        eq(_MULTICALL_ADDRESS, sender)
+                        and(lt(0x03, data.length), iszero(shl(0x60, xor(MULTICALL_ADDRESS, sender))))
```
e1Ru1o, line 63: I am missing what is this `lt` for 🤔

src/core/EulerSwap.sol
```diff
@@ -0,0 +1,505 @@
...
+    {
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+
+            mstore(0x40, operator)
+            mstore(0x2c, shl(0x60, account)) // clears `operator`'s padding
+            mstore(0x0c, 0x1647292a000000000000000000000000) // selector for `isAccountOperatorAuthorized(address,address)` with `account`'s padding
+            if iszero(staticcall(gas(), evc, 0x1c, 0x44, 0x00, 0x20)) {
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
```
e1Ru1o, line 40: Not doing the new trick?
```suggestion
                let ptr_ := mload(0x40)
                returndatacopy(ptr_, 0x00, returndatasize())
                revert(ptr_, returndatasize())
```

e1Ru1o (commented):
Besides `findCurvePoint` and `calcLimits` everything seems ok. I need to do a second round on the 2 mentioned ones.

src/multicall/MultiCallContext.sol
duncancmt: If you check the implementation of `MultiCall`, you'll see that the sender is only forwarded if there's at least 4 bytes of data (to avoid confusion with the selector). This is somewhat redundant, given that we don't call this function in the `fallback` or `receive` paths, but it's good in case we want to reuse this later

src/core/EulerSwap.sol
duncancmt: `0x40` is clobbered above, so we can't reload it; we'll get an OOG

e1Ru1o: Ohh missed it, great.

e1Ru1o (approved):

src/core/UniswapV4.sol
```diff
@@ -298,7 +297,7 @@ abstract contract UniswapV4 is SettlerAbstract {
             }
             // TODO: price limits
             params.sqrtPriceLimitX96 = uint160(
-                zeroForOne.ternary(uint160(4295128740), uint160(1461446703485210103287273052203988822378723970341))
+                (!zeroForOne).ternary(uint160(1461446703485210103287273052203988822378723970341), uint160(4295128740))
```
e1Ru1o, line 300: is this an improvement? 😅  How so?

e1Ru1o (approved):

src/core/UniswapV4.sol
duncancmt: Yeah, it's actually quite a big improvement. It's due to how `ternary` is implemented and `solc`'s ability to optimize constant expressions. Basically, the opcodes emitted are the same, but the constants involved are shorter

test/unit/core/UniswapV3UnitTest.t.sol
```diff
@@ -253,9 +262,19 @@ contract UniswapV3UnitTest is Utils, Test {
...
+                        permitTransfer,
+                        hex"",
+                        minBuyAmount
+                    )
+                ),
+                address(this)
             ) // Forward on true msg.sender
         );
-        // uni.sell(RECIPIENT, data, minBuyAmount, permitTransfer, hex"");
+        // uni.sell(RECIPIENT, abi.encodePacked(TOKEN0, uint8(0), uint24(500), TOKEN1), minBuyAmount, permitTransfer, hex"");
```
e1Ru1o, line 278: uncomment?

src/core/UniswapV3Fork.sol
```diff
@@ -286,13 +284,12 @@ abstract contract UniswapV3Fork is SettlerAbstract {
...
         bytes32 salt;
         assembly ("memory-safe") {
-            token0 := and(ADDRESS_MASK, token0)
-            token1 := and(ADDRESS_MASK, token1)
             poolId := and(UINT24_MASK, poolId)
             let ptr := mload(0x40)
-            mstore(0x00, token0)
-            mstore(0x20, token1)
             mstore(0x40, poolId)
+            mstore(0x20, token1)
```
e1Ru1o, line 290: no masking it anymore?

test/unit/core/UniswapV3UnitTest.t.sol
duncancmt: It doesn't quite work because of how the padding with the strict ABIEncoding works. They're semantically equivalent, but the byte representations are different

src/core/UniswapV3Fork.sol
duncancmt: `token1` is masked by shifting `token0` up and then offsetting the address where we store `token0`. It costs 1 more gas, but it saves a bunch of contract size

PR #342, opened 2025-05-06
WIP: add a Settler variant for bridging

duncancmt:

PR #356, opened 2025-06-10
Bridge Settler

e1Ru1o:


src/allowanceholder/IAllowanceHolder.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.25;
 
+IAllowanceHolder constant ALLOWANCE_HOLDER = IAllowanceHolder(0x0000000000001fF3684f28c67538d4D072C22734);
+
```
duncancmt, line 5: Why?

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,58 @@
...
+            assert(IERC721Owner(DEPLOYER).ownerOf(_tokenId()) == address(this));
+        } else {
+            assert(gitCommit == bytes20(0));
+        }
+    }
+
+    function _dispatch(uint256, uint256 action, bytes calldata data) internal virtual override returns (bool) {
+        if (action == uint32(IBridgeSettlerActions.TAKE.selector)) {
+            (address token, uint256 amount) = abi.decode(data, (address, uint256));
+            ALLOWANCE_HOLDER.transferFrom(token, _msgSender(), address(this), amount);
```
duncancmt: This should be part of `Permit2Payment`

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,58 @@
...
+        }
+        else if (action == uint32(IBridgeSettlerActions.SETTLER_SWAP.selector)) {
+            (address token, uint256 amount, address settler, bytes memory settlerData) = abi.decode(data, (address, uint256, address, bytes));
+            
+            IERC20(token).approve(address(ALLOWANCE_HOLDER), amount);
```
duncancmt: Use `safeApproveIfBelow`

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    function _dispatch(uint256, uint256 action, bytes calldata data) internal virtual override returns (bool) {
+        if (action == uint32(IBridgeSettlerActions.TAKE.selector)) {
+            (address token, uint256 amount) = abi.decode(data, (address, uint256));
+            ALLOWANCE_HOLDER.transferFrom(token, _msgSender(), address(this), amount);
+        }
+        else if (action == uint32(IBridgeSettlerActions.SETTLER_SWAP.selector)) {
+            (address token, uint256 amount, address settler, bytes memory settlerData) = abi.decode(data, (address, uint256, address, bytes));
```
duncancmt, line 56: Consider the implications of passing `settler` in calldata vs validating it against `DEPLOYER.ownerOf(...)`

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,58 @@
...
+                token,
+                amount,
+                payable(settler),
+                settlerData
+            );
+        } else if (action == uint32(IBridgeSettlerActions.BRIDGE.selector)) {
+            (address token, address bridge, bytes memory bridgeData) = abi.decode(data, (address, address, bytes));
+
+            uint256 balance = IERC20(token).fastBalanceOf(address(this));
+            IERC20(token).approve(bridge, balance);
```
duncancmt: `safeApproveIfBelow`

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,58 @@
...
+            (bool success, ) = bridge.call(bridgeData);
+            if (!success) return false;
```
duncancmt: Why not bubble the revert? This is going to produce some pretty weird behavior. There's also a gas-optimized library for this pattern in `utils/Revert.sol`

src/bridge/BridgeSettler.sol
```diff
@@ -0,0 +1,39 @@
...
+import {revertActionInvalid} from "../core/SettlerErrors.sol";
+import {CalldataDecoder} from "../SettlerBase.sol";
+import {UnsafeMath} from "../utils/UnsafeMath.sol";
+
+interface IBridgeSettlerTakerSubmitted {
+    function execute(bytes[] calldata, bytes32)
+        external
+        payable
+        returns (bool);
+}
```
duncancmt, line 17: Why not reuse the existing interface?

src/bridge/BridgeSettler.sol
```diff
@@ -0,0 +1,39 @@
...
+        payable
+        override
+        returns (bool)
+    {
+        for (uint256 i = 0; i < actions.length; i = i.unsafeInc()) {
+            (uint256 action, bytes calldata data) = actions.decodeCall(i);
+            if (!_dispatch(i, action, data)) {
+                revertActionInvalid(i, action, data);
+            }
+        }
```
duncancmt: Consider carefully the security implications of not forcing the first action to be a VIP

src/bridge/BridgeSettlerAbstract.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+abstract contract BridgeSettlerAbstract {
+    // Do we need this? Is it going to be registered in the Registry?
+    // maybe take out logic related to this from Settler into a different contract
+    // and inherit from that everywhere
+    function _tokenId() internal pure virtual returns (uint256);
```
duncancmt: It's already abstracted in `SettlerAbstract`

duncancmt: If there's some more functionality you want to DRY, you can move that into SettlerAbstract too

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,58 @@
...
+    function _dispatch(uint256, uint256 action, bytes calldata data) internal virtual override returns (bool) {
+        if (action == uint32(IBridgeSettlerActions.TAKE.selector)) {
+            (address token, uint256 amount) = abi.decode(data, (address, uint256));
+            ALLOWANCE_HOLDER.transferFrom(token, _msgSender(), address(this), amount);
+        }
+        else if (action == uint32(IBridgeSettlerActions.SETTLER_SWAP.selector)) {
+            (address token, uint256 amount, address settler, bytes memory settlerData) = abi.decode(data, (address, uint256, address, bytes));
+            
+            IERC20(token).approve(address(ALLOWANCE_HOLDER), amount);
+            ALLOWANCE_HOLDER.exec(
```
duncancmt: I have no direct objection, but any time you interact with AllowanceHolder, it's a security-sensitive moment. Consider carefully what you're doing and leave a comment with your reasoning about why this is safe

test/unit/BridgeSettler.t.sol
```diff
@@ -0,0 +1,139 @@
...
+        vm.etch(address(USDC), address(token).code);
+        settler = new MainnetSettler(bytes20(0));
+    }
+
+    function testUserFlow() public {
+        address user = makeAddr("user");
+
+        bytes[] memory settlerActions = new bytes[](2);
+        // Take the assets from BridgeSettler
+        settlerActions[0] = abi.encodeWithSelector(
```
duncancmt: Using `abi.encodeCall` is mostly the same, but you also get type checking

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,58 @@
...
+                settler,
+                token,
+                amount,
+                payable(settler),
+                settlerData
+            );
+        } else if (action == uint32(IBridgeSettlerActions.BRIDGE.selector)) {
+            (address token, address bridge, bytes memory bridgeData) = abi.decode(data, (address, address, bytes));
```
duncancmt: Bug!!!! What happens if `bridge` is AllowanceHolder (or Permit2, but Permit2 isn't supported in this contract)

test/unit/BridgeSettler.t.sol
```diff
@@ -0,0 +1,139 @@
...
+                deadline: block.timestamp
+            }),
+            bytes(""),
+            abi.encodeWithSelector(
+                ALLOWANCE_HOLDER.transferFrom.selector,
+                address(token),
+                address(bridgeSettler),
+                address(settler),
+                1000
+            )
```
duncancmt: what

src/allowanceholder/IAllowanceHolder.sol
e1Ru1o: This was a change you did on the DirtyHack PR and as I was initially working on #342 I saw it and ended up doing it here. You put it in `AllowanceHolderContext.sol` file though. Why not 😅 ?

src/bridge/BridgeSettler.sol
e1Ru1o: For me BridgeSettler was going to follow Settler design but not necessarily by inheriting from it or using same structure. The existing interface has the AllowedSlippage argument and I don't see why BridgeSettler needs that. Slippage logic is relevant to Settler and I decided to abstract it away from the BridgeSettler execute.

src/bridge/BridgeSettlerAbstract.sol
e1Ru1o: I was feeling `SettlerAbstract` had more than what I wanted here, that was just the `_tokenId` logic and the constructor that checks the `_tokenId` ideally.

test/unit/BridgeSettler.t.sol
e1Ru1o: 🤦‍♂️ changed the action and left that unnecessary parameter there. Best way to learn to always use `encodeCall`

src/bridge/BridgeSettlerBase.sol
e1Ru1o: A bit lost here, I had a small discussion with you about this earlier, mentioning the options of taking Settler from the `DEPLOYER` (I was referring to it as registry 😅) vs getting it as parameter. You were in favor of taking it as parameter but if we decide to validate it that is same as getting it from DEPLOYER. 

Also, what happens with the dwell time?

I see the problem you are raising, I am allowing any address to take the assets which is most probably not inline with the defensive approach settler is implemented with. That is why I was considering to take it directly from DEPLOYER. 

But just realized that taking it from DEPLOYER doesn't give a chance to interact with prev during dwell time so, probably that is why you want it as argument, and we just need to check it to validate it is a proper active settler. Adding the check 😅

duncancmt: you don't necessarily need to check. but it would be good to justify the decision in a comment given that this is security sensitive. BridgeSettler itself doesn't have slippage checks, so we need to be confident that we're not introducing any new vulnerabilities

src/allowanceholder/IAllowanceHolder.sol
duncancmt: yeah, fair. it's not a problem, I was just wondering

src/bridge/BridgeSettlerBase.sol
```diff
@@ -5,18 +5,24 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
...
-abstract contract BridgeSettlerBase is AllowanceHolderContext, BridgeSettlerAbstract {
+abstract contract BridgeSettlerBase is SettlerAbstract {
     using SafeTransferLib for IERC20;
     using Revert for bool;
+    using FastDeployer for IDeployer;
 
     event GitCommit(bytes20 indexed);
 
+    IDeployer internal constant _DEPLOYER = IDeployer(DEPLOYER);
+    uint128 internal constant _SETTLER_TOKENID = 2;
```
duncancmt: I recommend renaming this constant to make it clear that this is the taker-submitted swap settler, and not some other settler variation

src/bridge/BridgeSettlerBase.sol
```diff
@@ -51,8 +51,13 @@ abstract contract BridgeSettlerBase is SettlerAbstract {
...
         if (action == uint32(IBridgeSettlerActions.SETTLER_SWAP.selector)) {
             (address token, uint256 amount, address settler, bytes memory settlerData) = abi.decode(data, (address, uint256, address, bytes));
-            
+            // Swaps are going to be directed to Settler, so `settler` must be an active settler
             _requireValidSettler(settler);
+            // To effectively do a swap we need to make funds accessible to Settler
+            // It is not possible to call it directly as the taker is going to be the BridgeSettler
+            // instead of the user, so, user assets needs to be pulled to BridgeSettler before
+            // attempting to do this swap.
+            // Security risks are inherited from Settler
```
duncancmt: Can you enumerate what risks you think we're inheriting?

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,94 @@
...
+                assembly ("memory-safe") {
+                    mstore(0x00, 0xe758b8d5) // selector for `ConfusedDeputy()`
+                    revert(0x1c, 0x04)
+                }
+            }
+
+            uint256 balance = IERC20(token).fastBalanceOf(address(this));
+            IERC20(token).safeApproveIfBelow(bridge, balance);
+            (bool success, bytes memory returnData) = bridge.call(bridgeData);
+            success.maybeRevert(returnData);
```
duncancmt: For my education, and for that of future developers, can you explain here why we don't use `basicSellToPool`? Is it just because we don't need `bps`? Because `basicSellToPool` has some features that I think we'll need.

e1Ru1o: Yep, it is basically because I feel `basicSellToPool` is out of place here. It has a very nice interphase to accommodate to the swapping interface while allowing some other use cases, but, here it feels over complicated for me atm. At least I am not convinced that is the best we can do.
We can think about the features and accommodate the implementation to it and if it is too similar to `basicSellToPool` fine we can use, but just reading "Pool" in SettlerBridge feels out of place for me (atm)

Also this is surely far from final implementation, I was just putting together something reasonable to start figuring out if we need something else or if what we had in mind works. Probably after working on some bridges we just figure out how the relative to `BASIC` in the bridge should look like.

src/bridge/BridgeSettlerBase.sol
e1Ru1o: Technically the only thing in my mind is that you can take away the assets if slippage is not set properly in settlerData and you do TRANSFER_FROM and BASIC to take the assets and send them respectively. 
Do you thing writing that there is relevant? I wrote that as a general comment that whatever you might be worried about Settler it applies here as well. I can probably write it in a less concerning way 😅.

Please suggest anything that you would like to add there. If there is a risk no-one knows better than you here.

src/bridge/BridgeSettlerBase.sol
e1Ru1o: After started working on integrating a bridge, it now makes sense to use the BASIC pattern here

src/bridge/BridgeSettlerBase.sol
```diff
@@ -0,0 +1,90 @@
...
+    function _requireValidSettler(address settler) private view {
+        // Any revert in `ownerOf` or `prev` will be bubbled. Any error in ABIDecoding the result
+        // will result in a revert without a reason string.
+        if (
+            _DEPLOYER.fastOwnerOf(_SETTLER_TAKER_SUBMITTED_TOKENID) != settler
+                && _DEPLOYER.fastPrev(_SETTLER_TAKER_SUBMITTED_TOKENID) != settler
+        ) {
+            assembly ("memory-safe") {
+                mstore(0x14, settler)
+                mstore(0x00, 0x7a1cd8fa000000000000000000000000) // selector for `CounterfeitSettler(address)` with `settler`'s padding
```
duncancmt, line 48: This selector must be declared otherwise it won't show up as part of the ABI

src/bridge/BridgeSettlerBase.sol
duncancmt: > if slippage is not set properly in settlerData and you do TRANSFER_FROM and BASIC to take the assets

yeah, please write that. I think that your assessment of the risks is correct.

duncancmt: footnote: did you decide how/when to handle wrapping ETH to WETH? seems like BridgeSettler isn't able to send ETH to Settler, so you'll need to wrap to WETH first before sending via AllowanceHolder.

duncancmt (approved):

src/bridge/BridgeSettlerBase.sol
e1Ru1o: Nice finding, let me check what to do there

e1Ru1o: I think I prefer to send the ETH to settler, I am just going to use the 0xeeeeee... address to decide what to send. Ok?

duncancmt: If you're sending ETH to Settler, you don't need to go through AllowanceHolder. you can just call `Settler.execute{value: ...}(...)` directly

src/bridge/BridgeSettlerBase.sol
```diff
@@ -19,6 +19,7 @@ abstract contract BridgeSettlerBase is Basic {
     using FastDeployer for IDeployer;
 
     event GitCommit(bytes20 indexed);
+    error CounterfeitSettler(address counterfeitSettler);
```
duncancmt, line 24: Consider adding this to `SettlerErrors.sol` or making a new file for BridgeSettler errors

src/bridge/BridgeSettlerBase.sol
```diff
@@ -58,7 +58,8 @@ abstract contract BridgeSettlerBase is Basic {
             // It is not possible to call it directly as the taker is going to be the BridgeSettler
             // instead of the user, so, user assets needs to be pulled to BridgeSettler before
             // attempting to do this swap.
-            // Security risks are inherited from Settler
+            // Settler can take over the assets if settlerData starts with TRANSFER_FROM which
+            // can put them at risk if AllowedSlippage is not properly set.
```
duncancmt: It's not just `TRANSFER_FROM`; it's any VIP action. And the risk here is specifically sandwiching from MEV attacks and forcing the swap to its slippage limit. Consider stating this explicitly.

src/bridge/BridgeSettlerBase.sol
e1Ru1o: yep yep doing so. Also considering if just making a separated action makes more sense / is more optimal?

e1Ru1o: oh wait, got distracted here, no change needed, they can do BASIC to wrap it and then whatever they need

src/bridge/BridgeSettlerBase.sol
e1Ru1o: Agreed in slack to keep it as it is. As it is used in assembly importing the error will not include it in the ABI

src/bridge/BridgeSettlerBase.sol
```diff
@@ -56,20 +56,31 @@ abstract contract BridgeSettlerBase is Basic, Relay, Mayan {
...
-                settler,
-                token,
-                amount,
-                payable(settler),
-                settlerData
-            );
+            if (token == address(ETH_ADDRESS)) {
+                // Native token swap
+                // Settler address was validated to be a correct settler, so we pass the
+                // arbitrary data as we know it is not a restricted target.
```
duncancmt, line 62: :+1:

duncancmt (approved):

PR #362, opened 2025-06-16
Relay Bridge Integration

e1Ru1o:
- Added graffiti in every action as an event with no parameters.
- Separated native asset bridge action from erc20 so less calldata is passed by avoiding the token 0xEEEE... to refer to ETH
- Requires #356 to be merged first

Notes on Relayer:
- Sending more or less than expected is not problematic, Relayers will take any amount sent and apply the quote rate set which will affect the received amount proportionally to the deposit
- There is not constraint on the sender

### DIFF
--- a/src/core/Relayer.sol
+++ b/src/core/Relayer.sol
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IERC20} from "@forge-std/interfaces/IERC20.sol";
+import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
+
+contract Relayer {
+    using SafeTransferLib for IERC20;
+
+    event RelayerAction(); // Graffiti for bridging operations through Relayer
+
+    function bridgeERC20ToRelayer(IERC20 token, address to, bytes32 requestId) internal {
+        emit RelayerAction();
+
+        uint256 amount = token.fastBalanceOf(address(this));
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            mstore(add(ptr, 0x54), requestId)
+            mstore(add(ptr, 0x34), amount)
+            mstore(add(ptr, 0x14), to)
+            mstore(ptr, 0xa9059cbb000000000000000000000000) // selector for `transfer(address,uint256)` with `to`'s padding
+
+            // Similar to SafeTransferLib.safeTransfer
+            if iszero(call(gas(), token, 0x00, add(0x10, ptr), 0x64, 0x00, 0x20)) { 
+                let ptr_ := mload(0x40)
+                returndatacopy(ptr_, 0x00, returndatasize())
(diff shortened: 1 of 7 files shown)


src/core/Relayer.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IERC20} from "@forge-std/interfaces/IERC20.sol";
+import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
+
+contract Relayer {
```
duncancmt: Is the name of the bridge `Relay` or `Relayer`?

src/core/Relayer.sol
```diff
@@ -0,0 +1,50 @@
...
+    using SafeTransferLib for IERC20;
+
+    event RelayerAction(); // Graffiti for bridging operations through Relayer
```
duncancmt: I think we probably need some more information in this event. Can you ask @Ktl-XV or @AndresElizondo ?

src/core/Relayer.sol
```diff
@@ -0,0 +1,50 @@
...
+
+    function bridgeERC20ToRelayer(IERC20 token, address to, bytes32 requestId) internal {
+        emit RelayerAction();
+
+        uint256 amount = token.fastBalanceOf(address(this));
+        assembly ("memory-safe") {
```
duncancmt: Can you think of a good way to DRY this with `SafeTransferLib`?

src/core/Relayer.sol
e1Ru1o: It is Relay 🤦‍♂️, correcting it

src/core/Relayer.sol
e1Ru1o: Checked with @pgrzesik but not with them, let's wait for their reply

src/core/Relayer.sol
e1Ru1o: I bring some code from there but I did not find a way to just use what is there already as there is no way to append the extra bytes

src/core/Relayer.sol
pgrzesik: Also @byeongminP - do you think it would be useful to add something e.g. Relay's `requestId` to this event or we're okay with getting it from the calldata?

For @sohai - do you think there's some piece of information that would be useful here for watching the events for milestone 2 with destination swaps? 

I think overall we only really care about the event itself as all the data is available from elsewhere, but maybe there's something that I'm missing

byeongminP: I think adding auxiliary indexed params to the event is not worth the extra gas cost in this case. Considering that we'd eventually want to cache quote params (intents, etc) for destination swaps, we can get away with storing the `requestId` extracted from the origin tx calldata off-chain.

e1Ru1o:
Merging as agreed today. We can do the event update once decided

PR #364, opened 2025-06-17
Mayan Bridge Integration

e1Ru1o:
- Interacts with `forwardEth` and `forwardERC20` in [Mayan Forwarder](https://github.com/mayan-finance/swap-bridge/blob/416affdf77239bd084a96535bdd3fb27ae52849d/src/MayanForwarder.sol#L164)

### DIFF
--- a/src/core/Mayan.sol
+++ b/src/core/Mayan.sol
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IERC20} from "@forge-std/interfaces/IERC20.sol";
+import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
+
+contract Mayan {
+    using SafeTransferLib for IERC20;
+
+    function bridgeERC20ToMayan(IERC20 token, address forwarder, address mayanProtocol, bytes memory protocolData) internal {
+        uint256 amount = token.fastBalanceOf(address(this));
+        token.safeApproveIfBelow(forwarder, amount);
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            
+            let size := mload(protocolData)
+            mcopy(add(0x134, ptr), protocolData, add(0x20, size))
+            mstore(add(0x114, ptr), 0x120)
+            mstore(add(0xf4, ptr), mayanProtocol)
+            // permit empty data as it is not going to be used
+            // zeroed to save gas
+            codecopy(add(0x54, ptr), codesize(), 0xa0)
+            // end of permit data
+            mstore(add(0x34, ptr), amount)
+            mstore(add(0x14, ptr), token)
+            // selector for `forwardERC20(address,uint256,(uint256,uint256,uint8,bytes32,bytes32),address,bytes)` with `token` padding
+            mstore(ptr, 0xe4269fc4000000000000000000000000) 
+
+            // modify copied amount in protocolData
+            // protocolData is (4 bytes selector, 32 bytes token, 32 bytes amount, ...anything else)
+            // it is stored at 0x134, so we need to skip size (0x20), selector (0x04) and token (0x20)
+            mstore(add(0x178, ptr), amount)
+
+            if iszero(call(gas(), forwarder, 0x0, add(0x10, ptr), add(0x144, size), 0x00, 0x00)) {
+                let ptr_ := mload(0x40)
+                returndatacopy(ptr_, 0x0, returndatasize())
+                revert(ptr_, returndatasize())
+            }
+        }   
+    }
+
+    function bridgeNativeToMayan(address forwarder, address mayanProtocol, bytes memory protocolData) internal {
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+
+            let size := mload(protocolData)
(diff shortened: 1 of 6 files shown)


src/core/Mayan.sol
```diff
@@ -0,0 +1,65 @@
...
+    function bridgeERC20ToMayan(IERC20 token, address forwarder, address mayanProtocol, bytes memory protocolData) internal {
+        uint256 amount = token.fastBalanceOf(address(this));
+        token.safeApproveIfBelow(forwarder, amount);
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            
+            let size := mload(protocolData)
+            mcopy(add(0x134, ptr), protocolData, add(0x20, size))
+            mstore(add(0x114, ptr), 0x120)
+            mstore(add(0xf4, ptr), mayanProtocol)
```
duncancmt: is `mayanProtocol` a constant?

duncancmt: doesn't `mayanProtocol` potentially have dirty bits?

src/core/Mayan.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IERC20} from "@forge-std/interfaces/IERC20.sol";
+import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
+
+contract Mayan {
+    using SafeTransferLib for IERC20;
+
+    function bridgeERC20ToMayan(IERC20 token, address forwarder, address mayanProtocol, bytes memory protocolData) internal {
```
duncancmt, line 10: Since `forwarder` is user-controlled, we need to either validate it against `isRestrictedTarget` or we need to add a comment explaining why this is harmless

src/core/Mayan.sol
```diff
@@ -0,0 +1,65 @@
...
+            mstore(ptr, 0xe4269fc4000000000000000000000000) 
+
+            // modify copied amount in protocolData
+            // protocolData is (4 bytes selector, 32 bytes token, 32 bytes amount, ...anything else)
+            // it is stored at 0x134, so we need to skip size (0x20), selector (0x04) and token (0x20)
+            mstore(add(0x178, ptr), amount)
+
+            if iszero(call(gas(), forwarder, 0x0, add(0x10, ptr), add(0x144, size), 0x00, 0x00)) {
+                let ptr_ := mload(0x40)
+                returndatacopy(ptr_, 0x0, returndatasize())
```
duncancmt:
```suggestion
                returndatacopy(ptr_, 0x00, returndatasize())
```

src/core/Mayan.sol
```diff
@@ -0,0 +1,65 @@
...
+                revert(ptr_, returndatasize())
+            }
+        }   
+    }
+
+    function bridgeNativeToMayan(address forwarder, address mayanProtocol, bytes memory protocolData) internal {
```
duncancmt, line 46: ditto previous comments about `forwarder` and `mayanProtocol`

src/core/Mayan.sol
```diff
@@ -0,0 +1,65 @@
...
+            mstore(ptr, 0xb0f584ff000000000000000000000000) // selector for `forwardEth(address,bytes)` with `mayanProtocol` padding
+
+            // modify copied amount in protocolData
+            // protocolData is (4 bytes selector, 32 bytes token, 32 bytes amount, ...anything else)
+            // it is stored at 0x54, so we need to skip size (0x20), selector (0x04) and token (0x20)
+            mstore(add(0x98, ptr), selfbalance())
+
+            if iszero(call(gas(), forwarder, selfbalance(), add(0x10, ptr), add(0x64, size), 0x00, 0x00)) {
+                let ptr_ := mload(0x40)
+                returndatacopy(ptr_, 0x0, returndatasize())
```
duncancmt:
```suggestion
                returndatacopy(ptr_, 0x00, returndatasize())
```

src/core/Mayan.sol
e1Ru1o: I think not 🤔, technically it can have but at the moment everything passed to functions on dispatch is decoded with `abi.decode` with validates the dirty bits

src/core/Mayan.sol
e1Ru1o: should be safe given that we don't do arbitrary calls to it.

src/core/Mayan.sol
e1Ru1o: And no it is not a constant

duncancmt: While `abi.decode` will definitely revert if the decoded data is not clean, there's nothing that says that its results always have their dirty bits cleaned. Of course, it would be kinda wild if they weren't, but the semantics of the compiler require each assembly block to handle dirty bits in its inputs

src/core/Mayan.sol
duncancmt: Ok, please leave a comment explaining the lack of arbitrary calls and explicitly stating the absence of any selector collisions

src/core/Mayan.sol
e1Ru1o: Masking it 🫡

PR #366, opened 2025-06-24
Stargate V2 bridge integration

e1Ru1o:
- for bridging ERC20, `token` can be taken from the pool, but after experimenting passing it as calldata is more gas efficient


src/bridge/BridgeSettlerBase.sol
```diff
@@ -110,6 +111,12 @@ abstract contract BridgeSettlerBase is Basic, Relay, Mayan, Across {
         } else if (action == uint32(IBridgeSettlerActions.BRIDGE_NATIVE_TO_ACROSS.selector)) {
             (address spoke, bytes memory depositData) = abi.decode(data, (address, bytes));
             bridgeNativeToAcross(spoke, depositData);
+        } else if (action == uint32(IBridgeSettlerActions.BRIDGE_ERC20_TO_STARGATE_V2.selector)) {
+            (address token, address pool, bytes memory sendData) = abi.decode(data, (address, address, bytes));
+            bridgeERC20ToStargateV2(IERC20(token), pool, sendData);
```
duncancmt:
```suggestion
            (IERC20 token, address pool, bytes memory sendData) = abi.decode(data, (IERC20, address, bytes));
            bridgeERC20ToStargateV2(token, pool, sendData);
```

test/integration/StargateV2.t.sol
```diff
@@ -0,0 +1,188 @@
...
+}
+
+contract StargateV2Test is BridgeSettlerIntegrationTest {
+    address pool;
+
+    receive() external payable {}
+
+    function _prepareSendToken(uint256 amount) internal returns (
+        IStargateV2.SendParam memory sendParam, 
+        IStargateV2.MessagingFee memory messagingFee,
```
duncancmt, line 65: It's generally a bad idea to return structs from functions. `solc` produces hideously inefficient code. It's much better to allocate the struct in the calling function and then mutate it in the helper function.

Of course, this is test code, so it doesn't really matter. You don't need to fix this. I'm adding this note here so that you know for the future.

e1Ru1o:
Running fmt in #367 to minimize the merge conflicts

PR #367, opened 2025-06-26
Bridges review / Improvements

e1Ru1o:


src/core/Mayan.sol
```diff
@@ -7,65 +7,64 @@ import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
 contract Mayan {
     using SafeTransferLib for IERC20;
 
-    function bridgeERC20ToMayan(IERC20 token, address forwarder, address mayanProtocol, bytes memory protocolData) internal {
+    function bridgeERC20ToMayan(address forwarder, bytes memory protocolAndData) internal {
+        IERC20 token;
+        assembly ("memory-safe") {
+            // protocolData is abi.encode(mayanProtocol, protocolData)
+            // and protocolData is (4 bytes selector, 32 bytes token, 32 bytes amount, ...anything else)
+            // then, token is at 0x84
```
duncancmt: What's the impact of this assumption being incorrect? Do we need to check for dirty bits?

duncancmt (commented):
What prompted these changes? Is there a failing test or PR in the internal repo that I can refer to?

src/core/Mayan.sol
```diff
@@ -7,65 +7,64 @@ import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
...
         assembly ("memory-safe") {
             let ptr := mload(0x40)
             
-            let size := mload(protocolData)
-            mcopy(add(0x134, ptr), protocolData, add(0x20, size))
-            mstore(add(0x114, ptr), 0x120)
-            mstore(add(0xf4, ptr), and(0xffffffffffffffffffffffffffffffffffffffff, mayanProtocol))
+            let size := mload(protocolAndData)
+            mstore(add(0x14, ptr), token)
+            // selector for `forwardERC20(address,uint256,(uint256,uint256,uint8,bytes32,bytes32),address,bytes)` with `token` padding
```
duncancmt, line 26: I'm a little confused how this struct in the middle is constructed. Is this the Permit data that is ignored?

e1Ru1o: Yes it is the permit data that is ignored

src/core/Mayan.sol
e1Ru1o: If it is not correct data is malformed and the underlying call is not going to work. `abi.encode(mayanProtocol, protocolData)` is the data (or part of it) needed in the function in the forwarder and we are passing it as it is. If there are dirty bits it is going to give an error down the line

PR #370, opened 2025-07-07
Deploy to HyperEVM

duncancmt:


src/chains/HyperEvm/Common.sol
```diff
@@ -0,0 +1,68 @@
...
+        override(/* SettlerAbstract, */ SettlerBase)
+        DANGEROUS_freeMemory
+        returns (bool)
+    {
+        if (super._dispatch(i, action, data)) {
+            return true;
+        } else {
+            return false;
+        }
+        return true;
```
e1Ru1o, line 43: I was about to ask why is this like that but guess it is just the template

e1Ru1o: Any plans of protocols to add here?

duncancmt: Yeah, it's just the template. There will be more things added as more protocols pop up. But right now, all the top DEX protocols I could find were just UniV3 forks, which are listed below

duncancmt:
merged

PR #372, opened 2025-07-21
EulerSwap solvency check

e1Ru1o:


src/core/EulerSwap.sol
```diff
@@ -593,3 +694,164 @@ abstract contract EulerSwap is SettlerAbstract {
...
+
+                if(zeroForOne) {
+                    newReserve0 += amount;
+                    newReserve1 -= amountOut;
+                } else {
+                    newReserve0 -= amountOut;
+                    newReserve1 += amount;
+                }
+
+                if(!CurveLib.verify(
```
duncancmt: Assuming that `findCurvePoint` is implemented correctly (which we can reasonably assume), then the only way for `verify` to fail is if a reserve goes over `type(uint112).max`. That's a much simpler check than this

src/core/EulerSwap.sol
```diff
@@ -593,3 +694,164 @@ abstract contract EulerSwap is SettlerAbstract {
...
+            }
+            
+            uint256 debt;
+            uint256 newDebt;
+            uint256 soldCollateral;
+            uint256 newCollateral;
+
+            // compute new debt in buyVault
+            {
+                uint256 collateralBalance = buyVault.fastBalanceOf(account);
```
duncancmt: This is not correct. `collateralBalance` is shares and `amountOut` is assets. You have to convert.

src/core/EulerSwap.sol
```diff
@@ -51,13 +51,50 @@ library FastEvc {
...
             }
             authorized := mload(0x00)
             mstore(0x40, ptr)
         }
     }
+
+    function fastGetCollaterals(IEVC evc, address account) internal view returns (address[] memory collaterals) {
+        assembly ("memory-safe") {
+            mstore(0x14, account)
+            mstore(0x00, 0xa4d25d1e000000000000000000000000) // selector for `getCollaterals(address)` with `evc`'s padding
```
duncancmt:
```suggestion
            mstore(0x00, 0xa4d25d1e000000000000000000000000) // selector for `getCollaterals(address)` with `account`'s padding
```

src/core/EulerSwap.sol
```diff
@@ -51,13 +51,50 @@ library FastEvc {
...
+            collaterals := mload(0x40)
+            returndatacopy(collaterals, 0x20, size)
+            mstore(0x40, add(collaterals, size))
+        }
+    }
+
+    function fastGetControllers(IEVC evc, address account) internal view returns (address[] memory controllers) {
+        assembly ("memory-safe") {
+            mstore(0x14, account)
+            mstore(0x00, 0xfd6046d7000000000000000000000000) // selector for `getControllers(address)` with `evc`'s padding
```
duncancmt:
```suggestion
            mstore(0x00, 0xfd6046d7000000000000000000000000) // selector for `getControllers(address)` with `account`'s padding
```

src/core/EulerSwap.sol
```diff
@@ -51,13 +51,50 @@ library FastEvc {
...
+            if iszero(staticcall(gas(), evc, 0x10, 0x24, 0x00, 0x00)) {
+                let ptr := mload(0x40)
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
+            }
+
+            let size := sub(returndatasize(), 0x20)
+            collaterals := mload(0x40)
+            returndatacopy(collaterals, 0x20, size)
+            mstore(0x40, add(collaterals, size))
```
duncancmt, line 77: :+1:

src/core/EulerSwap.sol
```diff
@@ -78,6 +115,26 @@ interface IOracle {
     // for computing collateral value, use `bidOutAmount`
 }
 
+library FastOracle {
+    function fastGetQuote(IOracle oracle, uint256 inAmount, IERC20 base, IERC20 quote) internal view returns (uint256 bidOutAmount, uint256 askOutAmount) {
```
duncancmt:
```suggestion
    function fastGetQuotes(IOracle oracle, uint256 inAmount, IERC20 base, IERC20 quote) internal view returns (uint256 bidOutAmount, uint256 askOutAmount) {
```

src/core/EulerSwap.sol
```diff
@@ -78,6 +115,26 @@ interface IOracle {
...
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+
+            mstore(ptr, 0x0579e61f) // selector for `getQuotes(uint256,address,address)`
+            mstore(add(0x20, ptr), inAmount)
+            mstore(add(0x40, ptr), and(0xffffffffffffffffffffffffffffffffffffffff, base))
+            mstore(add(0x60, ptr), and(0xffffffffffffffffffffffffffffffffffffffff, quote))
```
duncancmt: it is allowed to clobber the word beginning at `0x60`, as long as you restore it to `0x00` at the end of the assembly block. the first memory byte that is reserved is `0x80`

src/core/EulerSwap.sol
```diff
@@ -78,6 +115,26 @@ interface IOracle {
...
+            if iszero(staticcall(gas(), oracle, add(0x1c, ptr), 0x64, 0x00, 0x40)) {
+                let ptr_ := mload(0x40)
+                returndatacopy(ptr_, 0x00, returndatasize())
+                revert(ptr_, returndatasize())
+            }
+            bidOutAmount := mload(0x00)
+            askOutAmount := mload(0x20)
```
duncancmt, line 138: What's the reasoning behind omitting the returndata size check here, while including it for the `FastEvault` methods?

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+    ) internal view returns (bool) {
+        address[] memory collaterals = evc.fastGetCollaterals(account);
+        IEVault debtVault;
+        uint256 debt;
+
+        {
+            address[] memory controllers = evc.fastGetControllers(account);
+
+            if(controllers.length > 1) return false;
+            if(controllers.length == 1) {
```
duncancmt: Please run `forge fmt -- src/core/EulerSwap.sol`

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+                debtVault = IEVault(controllers[0]);
```
duncancmt: Indexing an array generates bloated/inefficient code. Consider:
```suggestion
                assembly ("memory-safe") {
                    debtVault := mload(add(0x20, controllers))
                }
```

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+        
+        uint256 newDebt;
+        uint256 soldCollateral;
+        uint256 newCollateral;
+
+        // compute new debt in buyVault
+        {
+            uint256 collateralBalance = buyVault.fastBalanceOf(account);
+            if(collateralBalance < amountOut) {
+                newDebt = amountOut - collateralBalance;
```
duncancmt: is it possible for this to underflow? also I think these are incompatible units. `amountOut` is in underlying, while `collateralBalance` is in vault shares

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+            }
+        }
+
+        // check sellVault debt
+        if(debtVault == sellVault) {
+            // debt repayment
+            if(amountIn < debt) {
+                // collateral in buyVault must be able to cover the amountOut
+                // to ensure there is only one controller at the end
+                if(newDebt != 0) return false;
```
duncancmt: Under what scenario can this conditional fail? Are there realistic conditions that could cause this?

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+                debt -= amountIn;
+            }
+            else {
+                newCollateral = amountIn - debt;
```
duncancmt: again, when can these underflow?

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+                debt = 0; // debt was repaid
+            }
+        } else {
+            newCollateral = amountIn;
+        }
+
+        // if there is new debt in buyVault, check that the controller is buyVault
+        // or debt was repaid to ensure there is only one controller at the end
+        if(newDebt != 0) {
+            if(debt != 0 && debtVault != buyVault) return false;
```
duncancmt: is this redundant with the earlier conditional I commented on?

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+            if(soldCollateral != 0) {
+                // adjust inequality to avoid substraction
+                // collateral + newCollateral - soldCollateral >= debt
+                // is changed to 
+                // collateral + newCollateral >= debt + soldCollateral
+                if(buyVault == debtVault) {
+                    debt += soldCollateral * 1e4;
+                } else {
+                    (uint256 value,) = oracle.fastGetQuote(soldCollateral, buyVault.fastAsset(), unitOfAccount);
+                    debt += (value * debtVault.fastLTVBorrow(buyVault));
```
duncancmt: Debt is not LTV-adjusted. Only collateral has LTV

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+                }
+            }
+            uint256 collateral;
+            if(newCollateral != 0) {
+                if(sellVault == debtVault) {
+                    collateral = newCollateral * 1e4;
```
duncancmt: you've lost me here. what is this doing?

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+                } else {
+                    (uint256 value,) = oracle.fastGetQuote(newCollateral, sellVault.fastAsset(), unitOfAccount);
+                    collateral = (value * debtVault.fastLTVBorrow(sellVault));
+                }
+            }
+            for(uint256 i = 0; i < collaterals.length; i++) {
+                IEVault collateralVault = IEVault(collaterals[i]);
+                (uint256 value,) = oracle.fastGetQuote(collateralVault.fastBalanceOf(account), collateralVault.fastAsset(), unitOfAccount);
```
duncancmt: You're mixing shares and underlying again here

src/core/EulerSwap.sol
```diff
@@ -592,4 +623,195 @@ abstract contract EulerSwap is SettlerAbstract {
...
+        (uint256 inLimit,) = EulerSwapLib.calcLimits(_EVC(), pool, zeroForOne, p, reserve0, reserve1);
+
+        uint256 sellAmount;
+        if (bps != 0) {
+            unchecked {
+                sellAmount = sellToken.fastBalanceOf(address(this)) * bps / BASIS;
+            }
+            // TODO: Comment for what happens with excess here.
+            // 1. Excess is absorbed by other sources
+            // 2. Ultimatelly donated as fee, which might result in a slippage revert? Double check.
```
duncancmt: please resolve TODO

src/core/EulerSwap.sol
e1Ru1o: Debt is in sell vault, buy vault is in the collaterals but running low because most of the debt is covered by other collaterals. If you try to deposit something less than the debt that consumes all the collateral in buy vault you are going to end up with two controllers. I don't see why that is not a valid case 🤔

src/core/EulerSwap.sol
e1Ru1o: If I provide both assets of the pool I understand I can borrow something else which might lead to a scenario were I ran out of collateral in one of the assets generating a new controller which is not allowed. If borrowing something else is not possible then you are right

src/core/EulerSwap.sol
duncancmt: I think the only way it's possible to achieve this is the following scenario: tokenA is in `vault0`, tokenB is in `vault1`, the pool is collateralized by tokenC (which may the same as tokenA or tokenB). After some trading has happened, the pool has incurred a debt in tokenA and credit in tokenB (`vault0` is the controller; `vault1` is enabled collateral). Then the account withdraws some of the credit in tokenB (either redeems or moves away some of the `vault1` shares). Now, when the pool gets back to equilibrium, both `vault0` and `vault1` will have debt and there will be 2 controllers when the EVC attempts to release its locks.

So yes, I agree that this is a condition that should be checked for and should throw an error. Can you make this comment explicit? Maybe annotate each of the variables with their semantics and try to give them consistent, evocative names so that it's clear what's going on?

src/core/EulerSwap.sol
duncancmt: I see. This is a possible scenario. It's definitely an invalid configuration that we should detect. Please make it explicit in the comments that this is what you're checking for.

src/core/EulerSwap.sol
e1Ru1o: Honestly I think that explaining all of that in the comment might be too much. The existing comment is true no matter the underlying case. If it is a valid check, the current comment should be enough as it is just focusing on the violation that needs to be avoided.

src/core/EulerSwap.sol
e1Ru1o: Same as in the other case applies here. This comment will be simpler but mentioning a possible scenario should not be needed as again the current comment is focusing on the controllers

src/core/EulerSwap.sol
duncancmt: There's no such thing as too much comments. Honestly, a good amount of comments in math-heavy code like this is probably 1:1 with code. Giving excessive amounts of context is probably a good thing because the next person to read this code will have much less context than we do.

src/core/EulerSwap.sol
duncancmt: Please add more comments. Having lots and lots of comments is a virtue, not a sin.

src/core/EulerSwap.sol
```diff
@@ -686,72 +739,82 @@ library EulerSwapLib {
...
+        // We now know the post-swap state of the pool. Adjust collateral for LTV and convert both
+        // collateral and debt into the unit of account for solvency.
         if (debt != 0) {
             IOracle oracle = debtVault.fastOracle();
             IERC20 unitOfAccount = debtVault.fastUnitOfAccount();
 
             (, debt) = oracle.fastGetQuotes(debt, debtVault.fastAsset(), unitOfAccount);
-            // multiply by 1e4 to avoid divisions when adjusting LTVBorrow bps of collaterals.
-            // All amounts in EulerSwap are lower than uint112 so there should not be overflow errors.
+            // Debt is not LTV adjusted. LTV is in basis points. By multiplying the debt bt 10_000,
```
e1Ru1o: typo, `bt 10_000`

PR #378, opened 2025-07-24
BridgeSettler deployment script

e1Ru1o:


sh/common_deploy_bridge_settler.sh
```diff
@@ -0,0 +1,73 @@
...
+declare deploy_bridge_settler_calldata
+deploy_bridge_settler_calldata="$(cast calldata "$deploy_sig" 5 "$bridge_settler_initcode")"
+declare -r deploy_bridge_settler_calldata
+
+if [[ -n "${deployer_address-}" ]] ; then
+    declare -a deploy_calldatas
+    if (( chainid == 534352 )) ; then
+        deploy_calldatas=(
+            0 "$deploy_bridge_settler_calldata" "$deployer_address"
+            1 "$(cast calldata "$multisend_sig" "$(cast concat-hex "${deploy_calldatas[@]}")")" "$multicall_address"
```
duncancmt: what is this delegatecall to the multicall doing? I don't see `deploy_calldatas` getting set beforehand anywhere, so it seems to me that this would just error out

sh/common_deploy_bridge_settler.sh
```diff
@@ -0,0 +1,73 @@
...
+        )
+    else
+        deploy_calldatas+=(
```
duncancmt: do you anticipate more than 1 flavor of bridge Settler? if not, then we probably don't need to be using multicall here. we can just send directly to the deployer

sh/confirm_new_contract.sh
```diff
@@ -121,14 +121,17 @@ cd "$project_root"
...
+
 declare safe_address
 safe_address="$(get_config governance.deploymentSafe)"
 declare -r safe_address
 
 . "$project_root"/sh/common_safe.sh
 . "$project_root"/sh/common_safe_owner.sh
 . "$project_root"/sh/common_wallet_type.sh
-. "$project_root"/sh/common_deploy_settler.sh
+. "$project_root"/sh/common_deploy_"$target_contract".sh
```
duncancmt: oooooooh.... this is extremely dangerous. definitely validate that `"$target_contract"` is an expected value and consider avoiding parameter expansion in `source` statements altogether

sh/deploy_new_contract.sh
```diff
@@ -139,7 +142,7 @@ if [[ $wallet_type != 'frame' ]] ; then
 fi
 
 . "$project_root"/sh/common_safe_deployer.sh
-. "$project_root"/sh/common_deploy_settler.sh
+. "$project_root"/sh/common_deploy_"$target_contract".sh
```
duncancmt: ditto previous comment

sh/deploy_new_contract.sh
```diff
@@ -296,4 +299,4 @@ while (( ${#deploy_calldatas[@]} >= 3 )) ; do
     SAFE_NONCE_INCREMENT=$((${SAFE_NONCE_INCREMENT:-0} + 1))
 done
 
-echo 'Contracts deployed. Run `sh/verify_settler.sh '"$chain_name"'` to verify on Etherscan.' >&2
+echo 'Contracts deployed. Run `sh/verify_'$target_contract'.sh '"$chain_name"'` to verify on Etherscan.' >&2
```
duncancmt: parameter expansion doesn't happen inside single quotes `'`. you have to use double quotes `"` or some other form of escape

sh/confirm_new_contract.sh
e1Ru1o: now taht you mention it, yes, makes sense that it is indeed dangerous and probably a bad practice. Will think about how to do it in a better way

sh/common_deploy_bridge_settler.sh
```diff
@@ -0,0 +1,37 @@
...
+bridge_settler_initcode="$(cast concat-hex "$(jq -Mr .bytecode.object < "$bridge_settler_artifact")" "$constructor_args")"
+declare -r bridge_settler_initcode
+
+declare -r deploy_sig='deploy(uint128,bytes)(address,uint32)'
+
+declare deploy_bridge_settler_calldata
+deploy_bridge_settler_calldata="$(cast calldata "$deploy_sig" 5 "$bridge_settler_initcode")"
+declare -r deploy_bridge_settler_calldata
+
+if [[ -n "${deployer_address-}" ]] ; then
```
duncancmt: what happens if `deployer_address` is unset?

sh/deploy_new_bridge_settler.sh
```diff
@@ -0,0 +1,299 @@
...
+declare signer
+IFS='' read -p 'What address will you submit with?: ' -e -r -i 0xEf37aD2BACD70119F141140f7B5E46Cd53a65fc4 signer
+declare -r signer
+
+. "$project_root"/sh/common_wallet_type.sh
+
+if [[ $wallet_type != 'frame' ]] ; then
+    echo 'This script only works with Frame.sh wallets due to the overly long calldata' >&2
+    exit 1
+fi
```
duncancmt: this is probably not true for bridge settler

sh/deploy_new_bridge_settler.sh
```diff
@@ -0,0 +1,299 @@
...
+    deploy_calldata_length="$(cast to-uint256 $deploy_calldata_length)"
+    deploy_calldata_length="${deploy_calldata_length:2}"
+
+    ## assemble the call to `execTransaction`
+    # we have to do it this awkward way instead of using `cast calldata` because
+    # `$deploy_calldata` is longer than allowed for a command-line argument
+    declare packed_calldata
+    packed_calldata="$(cast concat-hex "$execTransaction_selector" "$(cast to-uint256 "$target")" "$(cast to-uint256 0)" "$(cast to-uint256 320)" "$(cast to-uint256 $operation)" "$(cast to-uint256 0)" "$(cast to-uint256 0)" "$(cast to-uint256 0)" "$(cast to-uint256 "$(cast address-zero)")" "$(cast to-uint256 "$(cast address-zero)")" "$(cast to-uint256 $((320 + 32 + ${#deploy_calldata} / 2)))")""$deploy_calldata_length""$deploy_calldata""$packed_signatures_length""$packed_signatures"
+
+    ## set gas limit and add multiplier/headroom (again mostly for Arbitrum)
```
duncancmt: all of the direct manipulation of the RPC using `jq` and `curl` from here down is not needed if you assume that the calldata is not so excessively long that it can't be passed as a command-line argument. of course, if you don't care to simplify this, you might want to reify this pattern so that you're not having tons of code duplication with the existing script

sh/common_deploy_bridge_settler.sh
e1Ru1o: good one, I was trying to follow what `common_deploy_bridge_settler` does but it probably doesn't make sense here. It seems that if is for deploy new chain script. But well I am just going to fail if unset

e1Ru1o: I don't know, I am going to just remove the `if` for now, there is no way of executing this script without `deployer_address` as all the other scripts that depend on this also depend on `common_safe` which sets `deployer_address`

sh/deploy_new_bridge_settler.sh
e1Ru1o: True, right that you had this issue with Settler after it got too long

sh/deploy_new_bridge_settler.sh
e1Ru1o: Reverted the changes in this commit 2b586d6 where the manipulation was introduced. I mean, reverted it only in the `deploy_new_bridge_settler.sh` script which contains those changes as they were taken from `deploy_new_settler.sh`

sh/common_deploy_bridge_settler.sh
```diff
@@ -0,0 +1,36 @@
...
+
+declare deploy_bridge_settler_calldata
+deploy_bridge_settler_calldata="$(cast calldata "$deploy_sig" 5 "$bridge_settler_initcode")"
+declare -r deploy_bridge_settler_calldata
+
+declare -a deploy_calldatas
+deploy_calldatas=(
+    0 "$deploy_bridge_settler_calldata" "$deployer_address"
+)
+
```
duncancmt: nit: trim trailing whitespace

PR #379, opened 2025-07-25
Bridge settlers for each chain

e1Ru1o:


test/integration/Mayan.t.sol
```diff
@@ -19,6 +20,10 @@ contract MayanTest is BridgeSettlerIntegrationTest {
     address forwarder = 0x337685fdaB40D39bd02028545a4FfA7D287cC3E2;
     address mayanProtocol;
 
+    function testBridgeSettler() internal override {
+        bridgeSettler = new ArbitrumBridgeSettler(bytes20(0));
+    }
+
```
duncancmt: What is the point of this function? The EVM environment is reset between tests, so all that this is really testing for is the fact that `ArbitrumBridgeSettler`'s constructor doesn't revert

EDIT: I see that this function is `internal` and that you call it from `setUp()`. consider naming this function something different. it's confusing

src/chains/Blast/BridgeSettler.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import {SettlerAbstract} from "../../SettlerAbstract.sol";
+import {IBridgeSettlerActions} from "../../bridge/IBridgeSettlerActions.sol";
+import {BridgeSettler, BridgeSettlerBase} from "../../bridge/BridgeSettler.sol";
+import {Across} from "../../core/Across.sol";
+
+contract BlastBridgeSettler is BridgeSettler, Across {
+    constructor(bytes20 gitCommit) BridgeSettlerBase(gitCommit) {
+        assert(block.chainid == 81457 || block.chainid == 31337);
+        // TODO: Check if BLAST configs are needed here
```
duncancmt: They are needed. Please add them

src/chains/Mode/BridgeSettler.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import {SettlerAbstract} from "../../SettlerAbstract.sol";
+import {IBridgeSettlerActions} from "../../bridge/IBridgeSettlerActions.sol";
+import {BridgeSettler, BridgeSettlerBase} from "../../bridge/BridgeSettler.sol";
+import {Across} from "../../core/Across.sol";
+
+contract ModeBridgeSettler is BridgeSettler, Across {
+    constructor(bytes20 gitCommit) BridgeSettlerBase(gitCommit) {
+        assert(block.chainid == 34443 || block.chainid == 31337);
+        // TODO: Check if MODE configs are needed here
```
duncancmt: ditto previous comment

duncancmt (approved):
looks good. please run the formatter.

test/integration/Mayan.t.sol
e1Ru1o: umm this was a patter we used before, like `testBlock`, but I definitely see how it is confusing. Will consider other naming options, maybe just start with underscore clear it out

duncancmt: well shame on me for establishing that pattern :sweat_smile: a leading underscore is probably fine

PR #380, opened 2025-08-01
Deploy Settler to Katana Network

duncancmt:


sh/common.sh
```diff
@@ -5,13 +5,9 @@ if ! hash forge &>/dev/null ; then
...
-declare forge_version
-forge_version="$(forge --version)"
-forge_version="${forge_version:13:7}"
-declare -r forge_version
-if [[ $forge_version != '59f354c' ]] ; then
-    echo 'Wrong foundry version installed -- '"$forge_version" >&2
-    echo 'Run `foundryup -i nightly-59f354c179f4e7f6d7292acb3d068815c79286d1`' >&2
+if [[ $(forge --version) != *b918f9b4ab0616b44e660a6bf8c5a47feece6505* ]] ; then
+    echo 'Wrong foundry version installed' >&2
+    echo 'Run `foundryup -i v1.3.0`' >&2
```
e1Ru1o, line 10: any comments about this switch? How did you decide to do so?

duncancmt: It was stable enough to make the deployment, and it has some improvements, particularly around the issue that I encountered here foundry-rs/foundry#11186

PR #381, opened 2025-08-04
Prepare for Osaka hardfork

duncancmt:


duncancmt:
See also #424 and #416

test/integration/PancakeInfinity.t.sol
e1Ru1o, line 1: what are the `sqrtPriceLimitX96` changes here for? 🤔

test/integration/SettlerBasePairTest.t.sol
```diff
@@ -57,7 +56,7 @@ abstract contract SettlerBasePairTest is BasePairTest {
         vm.chainId(31337);
         settler = _deploySettler();
         vm.label(address(settler), "Settler");
-        vm.etch(address(allowanceHolder), address(new AllowanceHolder()).code);
+        vm.etch(address(allowanceHolder), vm.getDeployedCode("AllowanceHolder.sol:AllowanceHolder"));
```
e1Ru1o, line 59: What motivated this change? Just for me to know

test/integration/SettlerMetaTxnPairTest.t.sol
```diff
@@ -45,6 +44,8 @@ abstract contract SettlerMetaTxnPairTest is SettlerBasePairTest {
 
         uint256 forkChainId = (new Shim()).chainId();
         vm.chainId(31337);
+        // Preserve the settlerMetaTxn address for the hardcoded signing hash.
+        new NonceBump();
```
e1Ru1o, line 48: This is because of the removal of `new AllowanceHolder()` if I followed properly, is it too complex to just change the harcoded values? I would do that or modify the comment a bit, the "Preserve" makes sense now that you are doing the change but later on I think it is easy to forget why it needed to be preserved, I suggest just explaining that it needs to be bumped and maybe mention what is the expected nonce

test/integration/PancakeInfinity.t.sol
duncancmt: The integration tests OOG with the Osaka-hardfork gas limit on the various tick-based AMMs we test against. Adding a meaningful `sqrtPriceLimitX96` fixes this so that the tests can pass. It's infuriating.

test/integration/SettlerBasePairTest.t.sol
duncancmt: AllowanceHolder builds with 0.8.25 to make sure it's the same everywhere. Now that we're moving to Osaka, we have to compile most of the contracts with 0.8.33. In order to keep AllowanceHolder on the old solc version, we have to access it as an artifact rather than compiling it into the test contract.

e1Ru1o (approved):
I saw CI was not passing, other than that it looks good

test/integration/SettlerBasePairTest.t.sol
e1Ru1o: Yep, good, thought so by the end of the review, makes total sense

test/integration/SettlerMetaTxnPairTest.t.sol
duncancmt: Correct.

We don't want to change the hardcoded value in test/integration/SettlerMetaTxnPairTest.t.sol (look for the giant warning) because it corresponds to an integration test elsewhere in 0x's repos. If we change the nonce of the test contract, we have to change the hardcoded hash. It's laborious to coordinate that update, so I would rather just bump the nonce here as a kludge. Do you have a suggestion for what the comment should say?

If we forget to preserve the nonce, there's a CI test that will fail. And if you paper that over while ignoring the giant warning, then whomever does that is beyond help :sweat:

e1Ru1o: Got it. For me the comment should be something like a warning too
```solidity
// Account nonce is relevant for upcoming settlerMetaTxn deployment to preserve 
// the address used in hardcoded signing hash. Expected nonce is X. 
// Dummy deployment to bump nonce to X
```

PR #382, opened 2025-08-05
Add Katana and HyperEVM bridge settlers

e1Ru1o:


duncancmt (approved):
What's going on with the submodule update here? Otherwise, looks good!

duncancmt (approved):

PR #390, opened 2025-08-13
LayerZeroOFT integration

e1Ru1o:
Added the action to ~~the chains supported by USDT0 only~~ all our chains

### DIFF
--- a/src/core/LayerZeroOFT.sol
+++ b/src/core/LayerZeroOFT.sol
@@ -3,11 +3,9 @@ pragma solidity ^0.8.25;
 
 import {IERC20} from "@forge-std/interfaces/IERC20.sol";
 import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
-import {FullMath} from "../vendor/FullMath.sol";
 
 contract LayerZeroOFT {
     using SafeTransferLib for IERC20;
-    using FullMath for uint256;
 
     function bridgeLayerZeroOFT(IERC20 token, address oft, bytes memory sendData) internal {
         uint256 updatedInputAmount = token.fastBalanceOf(address(this));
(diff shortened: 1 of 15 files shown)


src/core/LayerZeroOFT.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IERC20} from "@forge-std/interfaces/IERC20.sol";
+import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
+import {FullMath} from "../vendor/FullMath.sol";
+
+contract LayerZeroOFT {
+    using SafeTransferLib for IERC20;
+    using FullMath for uint256;
```
duncancmt: do we need `FullMath`?

test/integration/LayerZeroOFT.t.sol
```diff
@@ -0,0 +1,126 @@
...
+        token.safeApprove(address(ALLOWANCE_HOLDER), amount);
+
+        IOFT.SendParam memory sendParam = IOFT.SendParam({
+            dstEid: uint32(30110), // ARBITRUM
+            to: bytes32(uint256(uint160(makeAddr("recipient")))),
+            amountLD: amount,
+            minAmountLD: amount,
+            extraOptions: bytes(""),
+            composeMsg: bytes(""),
+            oftCmd: bytes("") // TAXI mode
```
duncancmt: what is TAXI mode?

e1Ru1o: It is not relevant here, Taxi mode was defined in Stargate but not in LayerZero, removing ...

PR #392, opened 2025-08-14
Support for SkyPSM

e1Ru1o:
Decided to do a breaking change instead of a separate action because the separate action approach was consuming too much contract size

### DIFF
--- a/src/core/MakerPSM.sol
+++ b/src/core/MakerPSM.sol
@@ -113,13 +113,13 @@ abstract contract MakerPSM is SettlerAbstract {
         USDC.safeApprove(address(SkyPSM), type(uint256).max);
     }
 
-    function sellToMakerPsm(address recipient, uint256 bps, bool buyGem, uint256 amountOutMin, IPSM psm, IERC20 gem)
+    function sellToMakerPsm(address recipient, uint256 bps, bool buyGem, uint256 amountOutMin, IPSM psm, IERC20 dai)
         internal
         returns (uint256 buyAmount)
     {
-        // If `psm/gem` is not `SkyPSM/USDS` or `LitePSM/DAI`, this interaction will likely fail
+        // If `psm/dai` is not `SkyPSM/USDS` or `LitePSM/DAI`, this interaction will likely fail
         // as those pairs are the ones with configured approvals in the constructor.
-        (IERC20 sellToken, IERC20 buyToken) = buyGem.maybeSwap(USDC, gem);
+        (IERC20 sellToken, IERC20 buyToken) = buyGem.maybeSwap(USDC, dai);
         uint256 sellAmount;
         unchecked {
             // phantom overflow can't happen here because:
@@ -136,7 +136,7 @@ abstract contract MakerPSM is SettlerAbstract {
                     revertTooMuchSlippage(buyToken, amountOutMin, buyAmount);
                 }
 
-                // gem.safeApproveIfBelow(address(psm), sellAmount);
+                // dai.safeApproveIfBelow(address(psm), sellAmount);
                 psm.fastBuyGem(recipient, buyAmount);
             }
         } else {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ Master list of UniV3 forks:
 ### Breaking changes
 
 * Update Scroll to the Cancun hardfork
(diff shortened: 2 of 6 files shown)


src/core/MakerPSM.sol
```diff
@@ -122,19 +143,19 @@ abstract contract MakerPSM is SettlerAbstract {
...
+                uint256 feeDivisor = psm.fastTout() + WAD; // eg. 1.001 * 10 ** 18 with 0.1% fee [tout is in wad];
                 // overflow can't happen at all because DAI is reasonable and PSM prohibits gemToken with decimals > 18
                 buyAmount = (sellAmount * USDC_basis).unsafeDiv(feeDivisor);
                 if (buyAmount < amountOutMin) {
                     revertTooMuchSlippage(buyToken, amountOutMin, buyAmount);
                 }
 
-                // DAI.safeApproveIfBelow(address(LitePSM), sellAmount);
-                LitePSM.fastBuyGem(recipient, buyAmount);
+                // DAI.safeApproveIfBelow(address(psm), sellAmount);
```
duncancmt: This comment is no longer relevant. Please remove it or update it.

src/core/MakerPSM.sol
```diff
@@ -104,15 +106,34 @@ abstract contract MakerPSM is SettlerAbstract {
...
+    }
+
+    function _sellToMakerPsm(address recipient, uint256 bps, bool buyGem, uint256 amountOutMin, IPSM psm, IERC20 asset)
+        private
+        returns (uint256 buyAmount)
+    {
+        (IERC20 sellToken, IERC20 buyToken) = buyGem.maybeSwap(USDC, asset);
         uint256 sellAmount;
         unchecked {
             // phantom overflow can't happen here because:
```
duncancmt, line 125: The line below also needs its comment updated

src/core/MakerPSM.sol
```diff
@@ -122,19 +143,19 @@ abstract contract MakerPSM is SettlerAbstract {
         }
         if (buyGem) {
             unchecked {
-                uint256 feeDivisor = LitePSM.fastTout() + WAD; // eg. 1.001 * 10 ** 18 with 0.1% fee [tout is in wad];
+                uint256 feeDivisor = psm.fastTout() + WAD; // eg. 1.001 * 10 ** 18 with 0.1% fee [tout is in wad];
                 // overflow can't happen at all because DAI is reasonable and PSM prohibits gemToken with decimals > 18
```
duncancmt:
```suggestion
                // overflow can't happen at all because DAI and USDS are reasonable and PSM prohibits gemToken with decimals > 18
```

src/core/MakerPSM.sol
```diff
@@ -104,15 +106,34 @@ abstract contract MakerPSM is SettlerAbstract {
...
-        (IERC20 sellToken, IERC20 buyToken) = buyGem.maybeSwap(USDC, DAI);
+        IPSM psm;
+        IERC20 asset;
+        if (sky) {
+            psm = SkyPSM;
+            asset = USDS;
+        } else {
+            psm = LitePSM;
+            asset = DAI;
+        }
```
duncancmt: Does this compile to the same bytecode if you use `sky ? ... : ...` ?

duncancmt (approved):
Can you check the impact on gas/code size of taking the asset/psm as arguments rather than hardcoding them and switching between them with the `sky` flag?

src/core/MakerPSM.sol
e1Ru1o: nope, doing the ternary twice consumes more contract size

duncancmt: Ahh, let me be more explicit:
```
(IPSM psm, IERC20 asset) = sky ? (SkyPSM, USDS) : (LitePSM, DAI);
```

e1Ru1o:
> Can you check the impact on gas/code size of taking the asset/psm as arguments rather than hardcoding them and switching between them with the `sky` flag?

Passing both as param it is a 100 units of gas increase based on the tests snapshots and around 100 bytes of contract size freed.

e1Ru1o:
> This looks good, but your nomenclature is incorrect. `gem` is USDC. It seems to me that USDS might be considered a `dai` in Maker parlance https://docs.makerdao.com/other-documentation/system-glossary

100% correct, changing it to dai to have the correct nomenclature

PR #396, opened 2025-08-19
Golf PancakeInfinity

e1Ru1o:


src/core/PancakeInfinity.sol
```diff
@@ -508,22 +507,27 @@ abstract contract PancakeInfinity is SettlerAbstract {
...
+                        IPancakeInfinityCLPoolManager.SwapParams({
+                            zeroForOne: zeroForOne,
+                            amountSpecified: amountSpecified,
+                            sqrtPriceLimitX96: uint160(
+                                zeroForOne.ternary(
+                                    uint160(4295128740), uint160(1461446703485210103287273052203988822378723970341)
+                                )
+                            )
+                        }),
+                        hookData
```
duncancmt: Constructing the `SwapParams` here causes a memory leak. Don't allocate inside a loop. Can you make the `SwapParams` members arguments to `unsafeSwap`? (also can you restore the removed TODO)

duncancmt (approved):

PR #398, opened 2025-08-20
Fix `"memory-safe"` violation; golf

duncancmt:


src/core/UniswapV2.sol
```diff
@@ -115,13 +116,13 @@ abstract contract UniswapV2 is SettlerAbstract {
...
             let sellAmountWithFee := mul(sellAmount, sub(10000, feeBps))
             buyAmount := div(mul(sellAmountWithFee, buyReserve), add(sellAmountWithFee, mul(sellReserve, 10000)))
-            let swapCalldata := add(ptr, 0x1c)
             // set up swap call selector and empty callback data
             mstore(ptr, UNI_PAIR_SWAP_SELECTOR)
             mstore(add(ptr, 0x80), 0x80) // offset to length of data
             mstore(add(ptr, 0xa0), 0) // length of data
 
             // set amount0Out and amount1Out
+            let swapCalldata := add(ptr, 0x1c)
```
e1Ru1o, line 125: All looks good, just wondering here why `add(ptr, 0x1c)` instead of `add(0x1c, pointer)`, is it to be consistent with the old code? or any other reason that I am missing?

duncancmt: This is just for consistency with the old code. I'll open a new PR to fix this

PR #401, opened 2025-08-23
DeBridge Integration

e1Ru1o:


CHANGELOG.md
```diff
@@ -48,6 +48,12 @@ Master list of UniV3 forks:
 
 ### Non-breaking changes
 
+* TODO:
```
duncancmt: I don't understand this change

src/core/DeBridge.sol
```diff
@@ -0,0 +1,100 @@
...
+        uint256 value;
+        assembly ("memory-safe") {
+            // offset to giveTokenAddress
+            inputToken := mload(add(0xe0, createOrderData))
+            value := selfbalance()
+        }
+        // Store the constant into source to read it only once
+        IDlnSource source = DLN_SOURCE;
+        if (address(inputToken) == address(0)) {
+            _bridgeNativeToDeBridge(source, value, value - globalFee, createOrderData);
```
duncancmt: is revert on underflow desired?

src/core/DeBridge.sol
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            uint256 amount = inputToken.fastBalanceOf(address(this));
+            inputToken.safeApproveIfBelow(address(source), amount);
+
+            _bridgeNativeToDeBridge(source, value, amount, createOrderData);
+        }
+    }
+
+    function _bridgeNativeToDeBridge(IDlnSource source, uint256 value, uint256 amount, bytes memory createOrderData)
+        internal
```
duncancmt:
```suggestion
        private
```

duncancmt (approved):

src/core/DeBridge.sol
e1Ru1o: I would say so, if `value` is below `globalFee` the txn is going to fail. How do you normally handle this case? Same thing in the `else` if there is no enough value to send the `globalFee` it will revert and it is not verified.

CHANGELOG.md
e1Ru1o: I am adding it as a TODO, to decide later on which chains to list once we do the release. Last release we deployed to a subset of the chains. It is just a note for later to update that properly.

duncancmt: Since these are unreleased changes, can you phrase this as-if we're doing the maximum deployment and we'll change it later when we cut the release if necessary

src/core/DeBridge.sol
duncancmt: If we don't have enough ETH, the `CALL` opcode will fail. I think we don't need to bother `Panic`ing here

e1Ru1o: We are always sending `value` that we have as it is the balance of the contract. That `value - globalFee` is the amount we want to bridge.

duncancmt: Ahh, OK. My bad. Then when we underflow, the amount that we're trying to bridge is negative. `Panic`ing is the appropriate action

PR #402, opened 2025-08-26
Allow trade surplus/positive slippage to be capped at a percentage of the trade

duncancmt:


src/SettlerBase.sol
```diff
@@ -142,14 +144,17 @@ abstract contract SettlerBase is ISettlerBase, Basic, RfqOrderSettlement, Uniswa
...
-            (address payable recipient, IERC20 token, uint256 expectedAmount) =
-                abi.decode(data, (address, IERC20, uint256));
+            (address payable recipient, IERC20 token, uint256 expectedAmount, uint256 maxBps) =
+                abi.decode(data, (address, IERC20, uint256, uint256));
             bool isETH = (token == ETH_ADDRESS);
             uint256 balance = isETH ? address(this).balance : token.fastBalanceOf(address(this));
             if (balance > expectedAmount) {
+                uint256 cap;
                 unchecked {
+                    cap = balance * maxBps / BASIS;
```
e1Ru1o, line 154: don't you want the `cap` to be in `expectedAmount` instead of the resulting amount?

e1Ru1o (approved):

src/SettlerBase.sol
duncancmt: the spec says that we want it to be capped at a proportion of the total settled amount. so it should be the resulting amount. after the cap is reached, the surplus amount continues to rise (slowly) with the buy amount

PR #404, opened 2025-09-11
Plasma chain

duncancmt:


api_secrets.json.template
```diff
@@ -86,5 +86,9 @@
   "katana": {
     "etherscanKey": "",
     "rpcUrl": ""
+  },
+  "plasma": {
+    "etherscanKey": "verifyContract",
```
duncancmt, line 91: This is weird, but it's correct. This is how you verify contracts on Routescan

PR #405, opened 2025-09-12
512-bit `sqrt`

duncancmt:


e1Ru1o (approved):
Math looks sound matching the comments

PR #419, opened 2025-10-15
chore: upload dao deployed safe addresses

gabririgo:
- upload dao safe addresses
- deployer is factory 0x4e1DCf7AD4e460CfD30791CCC4F9c8a4f820ec67
- singleton v1.4.1 0x41675C099F32341bf84BFc5382aF534df5C7461a
- monad-testnet market "NaN" as singleton v1.4.1 has not been deployed there (dao will maintain same safe address across chains)


duncancmt:
Oh, also, for chains that you do not intend to support, you can omit the key or set it to `null`. I'm not sure `"NaN"` will play nice with the deployment scripts, but I'll test it to make sure when we push this change live

PR #427, opened 2025-11-20
Add `MultiCallSendOnly` support to Safe{Wallet} guard

duncancmt:
In response to https://github.com/0xProject/0x-settler/pull/253#issuecomment-3208533193


src/deployer/SafeGuard.sol
```diff
@@ -0,0 +1,1172 @@
...
+        // end state (when the reentrancy lock is lifted) is valid.
+        {
+            uint256 ownerCount = _safe.ownerCount();
+            if (ownerCount < _MINIMUM_OWNERS) {
+                revert NotEnoughOwners(ownerCount);
+            }
+
+            uint256 threshold = _safe.getThreshold();
+            if (threshold < _MINIMUM_THRESHOLD) {
+                revert ThresholdTooLow(threshold);
```
jparklev, line 724: For a 3 of 5 safe, going through the resignation flow: 

`_getThresholdAfterResign` returns 3 -> `removeOwner` executes -> 3 of 4 -> this check hits (4 - 3 < 2) -> ThresholdTooHigh

Similar for 4 of 6, etc.

Is this intentional? I assume we want to keep at least some buffer of signers in case keys are lost or inaccessible, but this might block `removeOwner`s needed for `cancel` at common multisig configurations?

duncancmt: Let's consider the reverse? What if that scenario were to occur, that the safe had an owner removed and became 3-of-4? How does this affect the recovery game in both the "leaked key" and the "malicious signer" cases?

duncancmt: This is fixed now.

e1Ru1o (approved):

e1Ru1o:
Actually let's rename the Dedaub report name, it starts with Ox instead of 0x

jparklev (approved):
LGTM.

Re "...document that the initial configuration of the Safe should be 2/5 at minimum and in general more owners are better" (from Slack): the Safe (`0xf36b…2F51`) is still 2/4. Are we planning to add a 5th signer before installing the guard?

duncancmt:
> rename the Dedaub report name

Generally, I make absolutely no changes to the report file, as provided by the auditor. That includes the filename, even if it's a bit wonky.

> Are we planning to add a 5th signer before installing the guard?

Correct.

PR #511, opened 2026-02-26
Formally prove the correctness of Sqrt.sol and Cbrt.sol

duncancmt:


.github/workflows/sqrt512-formal.yml
```diff
@@ -0,0 +1,102 @@
...
+      - formal/sqrt/**
+      - formal/python/**
+      - test/0.8.25/formal-model/FormalModelFFI.t.sol
+      - test/0.8.25/formal-model/Sqrt512Model.t.sol
+      - .github/workflows/sqrt512-formal.yml
+  pull_request:
+    paths:
+      - src/utils/512Math.sol
+      - src/vendor/Sqrt.sol
+      - src/wrappers/Sqrt512Wrapper.sol
```
e1Ru1o, line 20: `SqrtWrapper` should be included as it is a dependency.
```suggestion
      - src/wrappers/SqrtWrapper.sol
      - src/wrappers/Sqrt512Wrapper.sol
      - formal/sqrt/**
      - formal/python/**
      - test/0.8.25/formal-model/FormalModelFFI.t.sol
      - test/0.8.25/formal-model/Sqrt512Model.t.sol
      - .github/workflows/sqrt512-formal.yml
  pull_request:
    paths:
      - src/utils/512Math.sol
      - src/vendor/Sqrt.sol
      - src/wrappers/SqrtWrapper.sol
      - src/wrappers/Sqrt512Wrapper.sol
```

formal/sqrt/SqrtProof/SqrtProof/FiniteCert.lean
e1Ru1o, line 1: It seems that this one got tracked before the gitignore was added.

formal/sqrt/SqrtProof/SqrtProof/FiniteCertSymbolic.lean
```diff
@@ -0,0 +1,623 @@
+import Init
+
+/-!
+Legacy symbolic certificate version (kept for reference).
```
e1Ru1o, line 4: are we keeping it?

.github/workflows/cbrt512-formal.yml
```diff
@@ -0,0 +1,101 @@
...
+        run: lake build && lake build cbrt512-model
+
+      - name: Check proof axioms
+        working-directory: formal/cbrt/Cbrt512Proof
+        run: |
+          printf '%s\n' \
+            'import Cbrt512Proof.GeneratedCbrt512Spec' \
+            'import Cbrt512Proof.CbrtUpSpec' \
+            '#print axioms Cbrt512Spec.model_cbrt512_evm_within_1ulp' \
+            '#print axioms Cbrt512Spec.model_cbrtUp512_wrapper_evm_correct' \
```
e1Ru1o, line 90: Maybe good to add the `Cbrt512Spec.model_cbrt512_wrapper_evm_correct` as well to the list of checked axioms depencies, technically to include the floor version of the cbrt512.

test/0.8.25/formal-model/Sqrt512Model.t.sol
e1Ru1o, line 1: Might be good to add a test case that forces overflow in the operations, similar to how it is done in `Sqrt.t.sol` to make sure it is checked.

Optionally which might be even better, consider extracting the relevant `sqrt` and `cbrt` from `Lib512MathTest` into a separated contracts that you can inherit in `Sqrt512Model` using the same override trick in `SqrtModel`, that way we avoid code duplication and we get the overflow case that is already implemented there plus any other like the existing regression test.

All of it applies as well to `Cbrt512Model.t.sol`

formal/sqrt/SqrtProof/SqrtProof/FiniteCertSymbolic.lean
duncancmt: no

formal/sqrt/SqrtProof/SqrtProof/FiniteCert.lean
duncancmt: yeah, this should be cleaned up by the time you get to #591

.github/workflows/sqrt512-formal.yml
duncancmt: This is fixed in #591

PR #552, opened 2026-04-05
Dry, review and comment bridges

e1Ru1o:


src/core/StargateV2.sol
```diff
@@ -4,36 +4,67 @@ pragma solidity ^0.8.25;
...
 
-    function bridgeNativeToStargateV2(address pool, uint256 destinationGas, bytes memory sendData) internal {
-        uint256 amount;
-        uint256 value;
+        // Get the native fee
         assembly ("memory-safe") {
-            value := selfbalance()
-            amount := sub(sub(value, destinationGas), mload(add(0x40, sendData)))
+            // sendData layout:
+            // +0x00: lenght of sendData
```
jparklev:
```suggestion
            // +0x00: length of sendData
```

src/core/Mayan.sol
```diff
@@ -4,60 +4,102 @@ pragma solidity ^0.8.25;
...
+            // +0x00: length of protocolAndData
+            // +0x20: mayanProtocol
+            // +0x40: offset to protocolData
+            let protocolAndDataPtr := add(0x20, protocolAndData)
+            let protocolDataOffsetPtr := add(0x20, protocolAndDataPtr)
+            let protocolDataOffset := mload(protocolDataOffsetPtr)
+            protocolDataPtr := add(protocolDataOffset, protocolAndDataPtr)
+            // Layout at `protocolDataPtr` is:
+            // +0x00: length of protocolData
+            // +0x04: selector to call in mayanProtocol
```
jparklev:
```suggestion
            // +0x00: length of protocolData (32 bytes)
            // +0x20: selector to call in mayanProtocol (4 bytes)
```

i believe?

test/unit/BridgeSettler.t.sol
```diff
@@ -45,6 +46,29 @@ contract BridgeSettlerTestBase is Test {
...
+
+    function _getSettlerDefaultTransferFrom(address token_, uint256 amount) internal returns (bytes memory) {
+        return _getDefaultTransferFrom(address(settler), token_, amount);
+    }
+
+    function _getDefaultTransferFrom(address settler_, address token_, uint256 amount) internal returns (bytes memory) {
+        return abi.encodeCall(
+            ISettlerActions.TRANSFER_FROM,
+            (
+                address(bridgeSettler),
```
jparklev:
```suggestion
                address(settler_),
```
(I believe this ensures tokens in `testUserFlow` do bridge settler -> settler -> bridge settler vs all bridge settler)

src/core/StargateV2.sol
```diff
@@ -4,36 +4,67 @@ pragma solidity ^0.8.25;
...
+            // +0x20: offset to SendParam
+            // +0x40: MessagingFee.nativeFee
+            // +0x60: MessagingFee.lzTokenFee
+            nativeFee := mload(add(0x40, sendData))
+        }
+
+        if (token == ETH) {
+            // Any excess on top to of fee + amount is returned to
```
jparklev:
```suggestion
            // Any excess on top of fee + amount is returned to
```

src/core/Mayan.sol
jparklev:
```suggestion
            // +0x00: length of protocolData
            // +0x20: selector to call in mayanProtocol
```

jparklev (approved):
Just nits. LGTM

src/core/Mayan.sol
e1Ru1o: Yes, it is like that, I think your way is more clear specifying the sizes

test/unit/BridgeSettler.t.sol
e1Ru1o: Good catch, need to modify the test to detect this error

jparklev (approved):

duncancmt (commented):
I have a general question: now that we are imposing a degree of structure on these `bytes` blobs that are passed around the various bridging actions, wouldn't it be both safer and less calldata-intensive to represent these fields explicitly (i.e. don't use `bytes` in the action definition) instead of as an offset in some somewhat-opaque blob? Seems like the present implementation will be a future source of bugs if we have any degree of drift in the encoding. `abi.decode` is actually pretty lightweight and it gives us a degree of surety that the encoding has been done correctly.

That's not to say that you have to do it this way, but it's something that you should consider as an alternative technique for the avoidance of bugs.

I still need to do another detailed pass over the specifics of the implementation. This is only a general comment.

src/core/StargateV2.sol
```diff
@@ -4,36 +4,67 @@ pragma solidity ^0.8.25;
...
+
+        if (token == ETH) {
+            // Any excess on top of fee + amount is returned to
+            // the refund address specified in `sendData`
+            uint256 value = address(this).balance;
+            // Input amount is adjusted with the pool convertRate
+            // any leftover contributes towards the refunded excess
+            updatedInputAmount = value - nativeFee;
+
+            nativeFee += updatedInputAmount;
```
duncancmt:
```suggestion
            nativeFee = value;
```

duncancmt (approved):

duncancmt (approved):

duncancmt (approved):

PR #558, opened 2026-04-15
chore: authorize feature 1001 to dao multisig on new chains

gabririgo:
This PR authorizes the DAO multisig a register a new feature when Settler is deployed on new chains.

- add 1001 as authorized DAO multisig feature for 365 days
- add deterministic DAO multisig address in config


script/SafeConfig.sol
```diff
@@ -133,4 +133,6 @@ library SafeConfig {
         result[5] = 0x36b7E0738fe11f05d26dA55d10eE679e684e06f4; // Lazaro
         return result;
     }
+
+    address internal constant daoSafe = 0x23030a6124E871F4744Cb9bc14D519b1f033FFe3;
```
duncancmt: This should not be hardcoded here, but should instead be derived from the toehold in use on the specific chain. On the one hand the Safe Singleton Factory toehold is not available on some chains (typically because they are not in general availability). On the other hand the Arachnid toehold isn't available on chains without non-EIP-155 transaction support. On the gripping hand the ERC-7955 toehold isn't available on chains that don't allow 7702-delegated accounts to `CREATE2`. Therefore, we have to be flexible.

sh/deploy_new_chain.sh
```diff
@@ -339,7 +339,8 @@ echo 'Add the following to your chain_config.json' >&2
 echo '"governance": {' >&2
 echo '	"upgradeSafe": "'"$upgrade_safe"'",' >&2
 echo '	"deploymentSafe": "'"$deployment_safe"'",' >&2
-echo '	"pause": "'"$ice_cold_coffee"'"' >&2
+echo '	"pause": "'"$ice_cold_coffee"'",' >&2
+echo '	"daoSafe": "0x23030a6124E871F4744Cb9bc14D519b1f033FFe3"' >&2
```
duncancmt: Ditto previous comment. This shouldn't be hardcoded here; it should be computed. Obviously, it's OK to hardcode in `chain_config.json` for those chains where deployment has already occurred.

script/DeploySafes.s.sol
```diff
@@ -549,11 +549,19 @@ contract DeploySafes is Script {
         );
         address predictedBridgeSettler = Create3.predict(salt(bridgeFeature, Nonce.wrap(1)), deployerProxy);
 
+        // DAO feature: initialize and authorize DAO Safe for feature 1001
+        Feature daoFeature = Feature.wrap(1001);
+        bytes memory daoSetDescriptionCall =
+            abi.encodeCall(Deployer.setDescription, (daoFeature, "0x DAO authorized feature"));
```
duncancmt: The description string here should be more descriptive. The current phrasing is in `sh/initial_description_crosschain_intent.md`. If you would like a different description for new chains (it's immutable on existing chains), please modify that file and also modify this script to ingest the contents of the file.

gabririgo: added a new `sh/initial_description_dao.md`  file because the exact feature is unknown atm.

sh/deploy_new_chain.sh
gabririgo: removed - address is computed with script-used modules

script/DeploySafes.s.sol
```diff
@@ -446,6 +447,23 @@ contract DeploySafes is Script {
...
+                safeFallback,
+                address(0),
+                0,
+                payable(address(0))
+            )
+        );
+        bytes32 daoSafeSalt =
+            keccak256(bytes.concat(keccak256(daoInitializer), bytes32(safeDeploymentSaltNonce)));
+
+        address daoSafe;
```
duncancmt: This should be passed into this Foundry script from the corresponding Bash script as an argument (the same as the other two Safes)

sh/initial_description_dao.md
```diff
@@ -0,0 +1,27 @@
...
+[Deployer](https://github.com/0xProject/0x-settler). Full source code and
+audits are found at
+[0xProject/0x-settler](https://github.com/0xProject/0x-settler).
+
+## Bug Bounty
+
+If you've found a bug, you may be eligible for a bounty under the 0x [bug bounty
+program](https://0x.org/docs/developer-resources/bounties). Please see that link
+for eligibility requirements as well as the advertised payout tiers. If you have
+any findings, send an email to [mailto:security@0x.org](security@0x.org) with
```
duncancmt:
```
[mailto:security@0x.org](security@0x.org)
```
should be
```
[security@0x.org](mailto:security@0x.org)
```

duncancmt (approved):
Very nice. Are you completely sure that the description markdown is exactly what you want it to be?

sh/deploy_new_chain.sh
```diff
@@ -131,6 +131,14 @@ if [[ ! -f "$project_root"/sh/initial_description_bridge_settler.md ]] ; then
     echo 'sh/initial_description_bridge_settler.md is missing' >&2
     exit 1
 fi
+if [[ ! -f "$project_root"/sh/initial_description_crosschain_intent.md ]] ; then
+    echo 'sh/initial_description_crosschain_intent.md is missing' >&2
+    exit 1
+fi
```
duncancmt, line 137: this appears to be unused

sh/initial_description_dao.md
gabririgo: removed the file and fell back to crosschain intents description for consistency, but fixed all existing descriptions files.

script/DeploySafes.s.sol
gabririgo: fixed now

sh/deploy_new_chain.sh
gabririgo: using this one for 1001 feature now, reverted addition of new description

PR #559, opened 2026-04-15
Clean up the handling of `bool` types in assembly for pedantic correctness

duncancmt:
https://docs.soliditylang.org/en/latest/internals/variable_cleanup.html


src/CrossChainReceiverFactory.sol
```diff
@@ -407,6 +407,8 @@ contract CrossChainReceiverFactory is ICrossChainReceiverFactory, MultiCallConte
         bytes32 proxyInitCode0 = _proxyInitCode0;
         bytes32 proxyInitCode1 = _proxyInitCode1;
         assembly ("memory-safe") {
+            setOwnerNotCleanup := iszero(iszero(setOwnerNotCleanup))
```
e1Ru1o, line 408: why not
```suggestion
            setOwnerNotCleanup := lt(0x00, setOwnerNotCleanup)
```
as everywhere else?

test/0.8.25/BoolBoundary.t.sol
```diff
@@ -0,0 +1,503 @@
...
+        MockEvcBool evc = new MockEvcBool();
+
+        evc.setResponse(new bytes(31));
+        vm.expectRevert(bytes(""));
+        harness.fastIsAuthorized(IEVC(address(evc)));
+
+        evc.setResponse(abi.encode(uint256(2)));
+        vm.expectRevert(bytes(""));
+        harness.fastIsAuthorized(IEVC(address(evc)));
+    }
```
e1Ru1o: These two don't seem very useful to me. `testFastEvcRejectsShortOrDirtyBool` is actually failing because we are not checking short data so, in this case, we read dirty data and there is no revert. I would remove these EVC tests

test/0.8.25/BoolBoundary.t.sol
```diff
@@ -0,0 +1,503 @@
...
+import {ISignatureTransfer} from "@permit2/interfaces/ISignatureTransfer.sol";
+
+import {FastLogic} from "src/utils/FastLogic.sol";
+import {Ternary} from "src/utils/Ternary.sol";
+import {UnsafeMath, Math} from "src/utils/UnsafeMath.sol";
+import {FastPermit} from "src/utils/SafePermit.sol";
+import {IDAIStylePermit} from "src/interfaces/IERC2612.sol";
+
+import {IEVC, FastEvc, IEulerSwap, FastEulerSwap} from "src/core/EulerSwap.sol";
+import {IUniV2Pair, fastUniswapV2Pool} from "src/core/UniswapV2.sol";
```
e1Ru1o: Can you rename this lib to upper case?
```suggestion
import {IUniV2Pair, FastUniswapV2Pool} from "src/core/UniswapV2.sol";
```

test/0.8.25/BoolBoundary.t.sol
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    function testEulerSwapBoundaryUsesCanonicalBit() public {
+        MockEulerSwap pool = new MockEulerSwap();
+
+        harness.fastEulerSwap(IEulerSwap(address(pool)), 9, address(0x55));
+        assertEq(pool.amount0Out(), 0);
+        assertEq(pool.amount1Out(), 9);
+        assertEq(pool.recipient(), address(0x55));
+    }
```
e1Ru1o, line 344: Example of the approach in the review comment using the Utlis lib with _mockExpectCall
```suggestion
    function testEulerSwapBoundaryUsesCanonicalBit() public {
        address pool = makeAddr("pool");

        _mockExpectCall(pool, abi.encodeCall(IEulerSwap.swap, (0, 9, address(0x55), bytes(""))),  bytes(""));
        harness.fastEulerSwap(IEulerSwap(pool), 9, address(0x55));
    }
```

test/0.8.25/BoolBoundary.t.sol
```diff
@@ -0,0 +1,503 @@
...
+
+    function testHanjiBoundaryCanonicalizesDirtyBool() public {
+        MockHanjiPool pool = new MockHanjiPool(address(0x11), address(0x22));
+
+        harness.hanjiPlaceMarketOrder(IHanjiPool(address(pool)));
+        assertTrue(pool.lastIsAsk());
+        assertTrue(
+            pool.lastSelector() == IHanjiPool.placeOrder.selector
+                || pool.lastSelector() == IHanjiPool.placeMarketOrderWithTargetValue.selector
+        );
```
e1Ru1o: This should not be an `||` here, this test matches `IHanjiPool.placeOrder`, a different call should be done for `IHanjiPool.placeMarketOrderWithTargetValue`

test/0.8.25/BoolBoundary.t.sol
```diff
@@ -0,0 +1,503 @@
...
+        assertEq(pool.lastData(), hex"abcd");
+    }
+
+    function testEkuboV2SwapCanonicalizesDirtyBool() public {
+        MockEkuboCore core = new MockEkuboCore();
+        EkuboPoolKey memory key = EkuboPoolKey({token0: address(0x11), token1: address(0x22), config: Config.wrap(0)});
+
+        harness.ekuboV2Swap(IEkuboCore(address(core)), key, 7, SqrtRatio.wrap(9));
+        assertTrue(core.lastIsToken1());
+    }
```
e1Ru1o, line 409: I think you should add a copy of this one for V3

test/0.8.25/BoolBoundary.t.sol
```diff
@@ -0,0 +1,503 @@
...
+
+    function fastUniswapV2GetReserves(address pool) external view returns (uint256 sellReserve, uint256 buyReserve) {
+        bool zeroForOne;
+        assembly ("memory-safe") {
+            zeroForOne := 0x02
+        }
+        return fastUniswapV2Pool.fastGetReserves(pool, zeroForOne);
+    }
+
+    function fastUniswapV2Swap(address pool, uint256 buyAmount, address recipient) external {
```
e1Ru1o, line 67: I think there is no harness for a dirty zeroForOne in UniV3, not even in the UniswapV3UnitTest

e1Ru1o (commented):
In general I wonder if it is better just to mock/expect calls instead of having all of this Mock contracts

Also in UniswapV3UnitTest.t.sol you did a test about isForwarded being dirty, I might want to do that for the other files with a similar modification, e.g EkuboV2, FlashAccounting

src/CrossChainReceiverFactory.sol
duncancmt: This contract is compiled for London, so the two expressions have the same gas, but `lt(0x00, ...)` is one byte longer.

src/CrossChainReceiverFactory.sol
```diff
@@ -950,7 +952,7 @@ contract CrossChainReceiverFactory is ICrossChainReceiverFactory, MultiCallConte
             if (address(this).balance < value) {
                 uint256 wrappedBalance;
                 IWrappedNative wnative = _WNATIVE;
-                bool hasWnative = _HAS_WNATIVE;
+                bool missingWnative = _MISSING_WNATIVE;
```
jparklev, line 953: Think `_HAS_WNATIVE` can be removed now, right?

test/unit/core/UniswapV3UnitTest.t.sol
```diff
@@ -277,4 +437,22 @@ contract UniswapV3UnitTest is Utils, Test {
...
+            abi.encode(abi.encodePacked(-int256(amount), -int256(amount))),
+            POOL
+        );
+
+        ISignatureTransfer.TokenPermissions memory permitted =
+            ISignatureTransfer.TokenPermissions({token: TOKEN0, amount: amount});
+        ISignatureTransfer.PermitTransferFrom memory permitTransfer =
+            ISignatureTransfer.PermitTransferFrom({permitted: permitted, nonce: 0, deadline: 0});
+
+        vm.expectRevert(bytes(""));
```
jparklev: I think this might be the cause of the foundry panic in ci (actually reverts with "Rejected" from the fallback dummy)

And if we had a proper permit2 mock here, it wouldn't reject, right? since the `0x02` is masked to `0x00` in https://github.com/0xProject/0x-settler/blob/b153a5a111b548cbe3c9b35346d262ac30190df7/src/core/UniswapV3Fork.sol#L368 (is this not intentional?)

test/0.8.25/BoolBoundary.t.sol
```diff
@@ -0,0 +1,503 @@
...
+        assertEq(y, 3);
+
+        assertEq(uint256(5).unsafeInc(dirtyTrue), 6);
+        assertEq(uint256(5).unsafeDec(dirtyTrue), 4);
+        assertEq(uint256(5).inc(dirtyTrue), 6);
+        assertEq(uint256(5).dec(dirtyTrue), 4);
+        assertEq(Math.toInt(dirtyTrue), 1);
+    }
+
+    function testFastDaiPermitCanonicalizesDirtyBool() public {
```
jparklev, line 251: Should we test the `!token.fastDAIPermit` path as well? with the new https://github.com/0xProject/0x-settler/blob/b153a5a111b548cbe3c9b35346d262ac30190df7/src/utils/SafePermit.sol#L275

jparklev (commented):
yay pendantic correctness, boo dirty bits. a few small things

e1Ru1o (approved):

PR #566, opened 2026-05-26
Add Nucleus Teller bridge support for WPAXG

e1Ru1o:
## Summary

Adds support for bridging WPAXG (Paxos Nucleus wrapped gold) through the Nucleus Teller contract via LayerZero v2. Includes two new bridge actions: `BRIDGE_TO_NUCLEUS_TELLER` for direct WPAXG bridging and `DEPOSIT_AND_BRIDGE_TO_NUCLEUS_TELLER` for wrapping PAXG into WPAXG before bridging.

## Key Changes

- **New `NucleusTeller` core mixin** (`src/core/NucleusTeller.sol`):
  - `bridgeToNucleusTeller()`: Bridges WPAXG shares held by the settler
  - `depositAndBridgeToNucleusTeller()`: Deposits PAXG, wraps to WPAXG, and bridges resulting shares
  - `_callTeller()`: Low-level assembly helper that forwards the settler's full ETH balance as bridge fees and handles LayerZero refunds
  - Dynamically overrides `shareAmount` and `depositAmount` parameters in calldata to use the settler's actual token balances

- **Updated bridge settler implementations**:
  - `MainnetBridgeSettler` and `OptimismBridgeSettler` now inherit from `NucleusTeller`
  - Added action dispatch handlers for both new bridge actions

- **Extended `IBridgeSettlerActions` interface**:
  - Added `BRIDGE_TO_NUCLEUS_TELLER` and `DEPOSIT_AND_BRIDGE_TO_NUCLEUS_TELLER` action selectors

- **Comprehensive integration tests** (`test/integration/NucleusTeller.t.sol`):
  - `testBridge_WPAXG_to_OP()`: Verifies WPAXG bridging with correct fee forwarding and share burning
  - `testBridge_WPAXG_to_OP_refundsExcessFee()`: Validates LayerZero fee refund handling
  - `testDepositAndBridge_PAXG_to_OP()`: Tests PAXG deposit, wrapping, and bridging flow

## Implementation Details

- Nucleus Teller address (`0xeE98730AAAdA5e6e092cA69F1AC1B9B554c059dF`) and WPAXG token address are hardcoded as they are identical across all supported chains
- Assembly-based calldata manipulation efficiently overrides amount parameters without full re-encoding
- Full ETH balance forwarding to Teller enables LayerZero to refund excess fees back to the settler
- Tests use fork testing at mainnet block 25180132 where both Teller and BoringVault are deployed

https://claude.ai/code/session_016x6xL6LehzfZsydng4KyHJ

### DIFF
--- a/src/core/NucleusTeller.sol
+++ b/src/core/NucleusTeller.sol
@@ -4,6 +4,11 @@ pragma solidity ^0.8.25;
 import {IERC20} from "@forge-std/interfaces/IERC20.sol";
 import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
 
+/// @dev Mirrors the relevant subset of paxoslabs/nucleus-boring-vault `AccountantWithRateProviders`.
+interface INucleusAccountant {
+    function getRateInQuoteSafe(IERC20 quote) external view returns (uint256);
+}
+
 /// @dev Mirrors the relevant subset of paxoslabs/nucleus-boring-vault `CrossChainTellerBase`.
 /// `BridgeData` follows `src/interfaces/ICrossChainTypes.sol` from that repo.
 interface INucleusTeller {
@@ -22,6 +27,10 @@ interface INucleusTeller {
         payable;
 
     function previewFee(uint256 shareAmount, BridgeData calldata data) external view returns (uint256);
+
+    function accountant() external view returns (INucleusAccountant);
+
+    function vault() external view returns (IERC20);
 }
 
 /// @title NucleusTeller
--- a/src/core/NucleusTeller.sol
+++ b/src/core/NucleusTeller.sol
@@ -4,13 +4,9 @@ pragma solidity ^0.8.25;
 import {IERC20} from "@forge-std/interfaces/IERC20.sol";
 import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
 
-/// @dev Mirrors the relevant subset of paxoslabs/nucleus-boring-vault `AccountantWithRateProviders`.
-interface INucleusAccountant {
-    function getRateInQuoteSafe(IERC20 quote) external view returns (uint256);
-}
-
-/// @dev Mirrors the relevant subset of paxoslabs/nucleus-boring-vault `CrossChainTellerBase`.
-/// `BridgeData` follows `src/interfaces/ICrossChainTypes.sol` from that repo.
+/// @dev Mirrors the relevant subset of paxoslabs/nucleus-boring-vault
+/// `MultiChainLayerZeroTellerWithMultiAssetSupport`. `BridgeData` follows
+/// `src/interfaces/ICrossChainTypes.sol` from that repo.
 interface INucleusTeller {
     struct BridgeData {
         uint32 chainSelector;
@@ -27,20 +23,16 @@ interface INucleusTeller {
         payable;
 
     function previewFee(uint256 shareAmount, BridgeData calldata data) external view returns (uint256);
-
-    function accountant() external view returns (INucleusAccountant);
-
-    function vault() external view returns (IERC20);
 }
 
 /// @title NucleusTeller
-/// @notice BridgeSettler integration for the Paxos Nucleus WPAXG `CrossChainTellerBase`.
+/// @notice BridgeSettler integration for the Paxos Nucleus WPAXG Teller.
 /// @dev The Teller and WPAXG share token addresses are identical on every chain the deployment
 /// supports (Ethereum and Optimism), so they are hardcoded.
 contract NucleusTeller {
     using SafeTransferLib for IERC20;
 
-    /// @notice Paxos Nucleus WPAXG `CrossChainTellerBase` (same address on Ethereum and Optimism)
+    /// @notice Paxos Nucleus WPAXG Teller (same address on Ethereum and Optimism)
     address internal constant NUCLEUS_TELLER = 0xeE98730AAAdA5e6e092cA69F1AC1B9B554c059dF;
 
     /// @notice WPAXG share token / Nucleus `BoringVault` (same address on Ethereum and Optimism)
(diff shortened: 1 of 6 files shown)


src/core/NucleusTeller.sol
```diff
@@ -0,0 +1,96 @@
...
+    function depositAndBridgeToNucleusTeller(bytes memory depositAndBridgeCallData) internal {
+        IERC20 depositAsset;
+        assembly ("memory-safe") {
+            // depositAndBridgeCallData layout in memory:
+            // +0x00: bytes length
+            // +0x20: depositAsset              <- read
+            // +0x40: depositAmount             <- override
+            // +0x60: minimumMint
+            // +0x80: offset to BridgeData tuple
+            depositAsset := mload(add(0x20, depositAndBridgeCallData))
```
duncancmt, line 68: What is the desired behavior if `depositAsset` is dirty?

src/core/NucleusTeller.sol
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        uint256 depositAmount = depositAsset.fastBalanceOf(address(this));
```
duncancmt, line 71: Are there a limited range of expected `depositAsset` values?

src/core/NucleusTeller.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IERC20} from "@forge-std/interfaces/IERC20.sol";
+import {SafeTransferLib} from "../vendor/SafeTransferLib.sol";
+
+/// @dev Mirrors the relevant subset of paxoslabs/nucleus-boring-vault
```
duncancmt: Please pin a specific commit in this comment or reference a deployed contract by its address

test/integration/NucleusTeller.t.sol
```diff
@@ -0,0 +1,182 @@
...
+    function testDepositAndBridge_PAXG_to_OP() public {
+        uint256 depositAmount = 1e18;
+
+        deal(address(PAXG), address(this), depositAmount, true);
+        PAXG.safeApprove(address(ALLOWANCE_HOLDER), depositAmount);
+
+        INucleusTeller.BridgeData memory data = _bridgeData();
+        // PAXG → WPAXG is a 1:1 wrap, so the resulting share count equals the deposit amount.
+        uint256 expectedShares = depositAmount;
+        uint256 fee = INucleusTeller(TELLER).previewFee(expectedShares, data);
```
duncancmt, line 154: What is `fee` expected to be? Is it zero?

duncancmt (approved):
Looks good!

pgrzesik (approved):
looks good from my perspective 👍

test/integration/NucleusTeller.t.sol
e1Ru1o: no, that is the fee for the bridging action (not the wrapping), which is the underlying LayerZero message fee. Definitely not zero.

src/core/NucleusTeller.sol
e1Ru1o: it is going to fail down the line, will add a comment

src/core/NucleusTeller.sol
e1Ru1o: a bit confused here, do you mean expected range of `depositAmount`? Or are you asking about the different deposit assets that can be used?

If you are asking about `depositAsset`, they only have PAXG enabled atm but they might enable others.

If you are asking about the amount (which I think you are), this is a clever question. While they don't have any limits for the incoming amount, they do have limits for the outgoing amount, defined as `minimunMint` in `depositAndBridge` function, then, we might want to scale that according to the new deposit amount. If the amount gets lowered and `minimunMint` has no margin for that decrement the transaction is going to revert.
Good catch 👌 

@pgrzesik for awareness that I am going to implement that change

e1Ru1o: After discussing with @pgrzesik we are dropping the scaling in favor of `minimumMint` acting as an strict slippage check for the route.

duncancmt: > If you are asking about `depositAsset`, they only have PAXG enabled atm but they might enable others.

This is what I meant, but I'm glad we had a creative misunderstanding.

Are you quite sure that you want this not to be scaled? I know that @jparklev was working on that exact feature for `BASIC`

duncancmt (approved):
In the future, please attribute commits correctly. You are committing as `claude`, which is confusing. And the commit messages include links that are not viewable by anyone except yourself.

While you're at it, you might want to set up commit signing and vigilant mode. This will probably be an organization-wide requirement in the near future. Obviously, do not give `claude` or `codex` access to your commit signing key.

src/core/NucleusTeller.sol
pgrzesik: What exactly would be scaled in BASIC? 

In this particular case, we're setting `minimumMint` on API side based on minimum expected output from swap (if there is one) or based on computed rate of PAXG->WPAXG (should be 1:1 but technically Accountant could report a different rate). Having this rescale doesn't change much about the guaratnees that we give the end user as we have to compute that value offchain anyway

e1Ru1o: We are fine not scaling it, the API can control what they want in there and force a revert if we go under that limit. That is effectively slippage for the PAXG->WPAXG wrapping and it doesn't affect the bridging.
For now I am convinced that there is no reason to scale it. If API team find a good reason to scale it, we can bring it back.

e1Ru1o:
> In the future, please attribute commits correctly. You are committing as `claude`, which is confusing. And the commit messages include links that are not viewable by anyone except yourself.

Will do, I was doing this PR in a remote claude code env, that is why claude is the one that did the commits.

PR #567, opened 2026-05-27
Renegade v2 integration

weijiekoh:
This PR upgrades the 0x Settler integration with Renegade (https://renegade.fi) v2.

Renegade v2 is deployed to Base Mainnet and Arbitrum One. On both deployments, the function signature for settlement is the same: `sponsorExternalMatch(uint256,address,(address,address,(uint256),uint256,uint256,uint256),(bool,uint8,bytes),(address,bool,uint256,uint256,bytes))`

The adapter contract exposes `sellToRenegade()` with the following signature. Unlike the v1 adapter contract, there is no `bool baseForQuote` parameter.

```
function sellToRenegade(
    address target,
    IERC20 sellToken,
    bytes memory data,
    uint256 minBuyAmt
) internal returns (uint256 buyAmt)
```

Renegade's GasSponsorV2 proxy contract addresses are hardcoded, Settler contract will need to be upgraded if Renegade deploys to a another chain.

### DIFF
--- a/src/ISettlerActions.sol
+++ b/src/ISettlerActions.sol
@@ -276,8 +276,7 @@ interface ISettlerActions {
         uint256 amountOutMin
     ) external;
 
-    function RENEGADE(address target, address sellToken, bool baseForQuote, bytes memory data, uint256 minBuyAmount)
-        external;
+    function RENEGADE(address target, address sellToken, bytes memory data, uint256 minBuyAmount) external;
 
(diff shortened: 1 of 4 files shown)


jparklev:
@weijiekoh Appreciate you putting this up, it's helpful and it'll speed us along

Picking up the work here, off of this commit, for our own internal coordination https://github.com/0xProject/0x-settler/pull/575

PR #575, opened 2026-06-01
Renegade v2

jparklev:
Picked up from https://github.com/0xProject/0x-settler/pull/567

Context from 567:
Renegade v2 is deployed to Base Mainnet and Arbitrum One. On both deployments, the function signature for settlement is the same: sponsorExternalMatch(uint256,address,(address,address,(uint256),uint256,uint256,uint256),(bool,uint8,bytes),(address,bool,uint256,uint256,bytes))

"The adapter contract exposes sellToRenegade() with the following signature. Unlike the v1 adapter contract, there is no bool baseForQuote parameter.

```
> function sellToRenegade(
    address target,
    IERC20 sellToken,
    bytes memory data,
    uint256 minBuyAmt
) internal returns (uint256 buyAmt)
```


src/core/Renegade.sol
```diff
@@ -5,110 +5,145 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
...
-            }
-        }
+    /// @notice The expected `GasSponsorV2` proxy address for the current chain.
+    /// @dev Adding a new chain requires a source change + redeploy of this contract.
+    function _renegadeGasSponsorV2() internal view returns (address) {
+        uint256 chainId = block.chainid;
+        if (chainId == 42161) return 0xcE7a8D45daa9a5B29f6d255552F577d53fF9EBcf; // Arbitrum One
+        if (chainId == 8453) return 0xD9E0507D706408D0f14E22e50880189Fd915be80; // Base mainnet
+        revert("Renegade: unsupported chain");
     }
```
duncancmt: You probably want this to be a `virtual` function without a default implementation that you override in the chain-specific mixin in order to avoid carrying chain-specific dead bytecode. This is relevant on Base.

src/core/Renegade.sol
```diff
@@ -5,110 +5,145 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
...
+    function sellToRenegade(address target, IERC20 sellToken, bytes memory data, uint256 minBuyAmt)
+        internal
+        returns (uint256 buyAmt)
+    {
+        // Extract the recipient from `data` and require that it resolves to address(this).
+        address recipient;
+        assembly ("memory-safe") {
+            recipient := mload(add(data, 0x40))
         }
+        require((recipient == address(0)) || (recipient == address(this)), "Renegade: bad recipient");
```
duncancmt: What does `address(0)` indicate?

src/core/Renegade.sol
```diff
@@ -5,110 +5,145 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
...
+        // Extract price from `data`.
+        uint256 priceRepr;
+        assembly ("memory-safe") {
+            priceRepr := mload(add(data, 0xa0))
         }
 
-        uint32 selector = _renegadeSelector();
+        // newBuyAmt = floor(newSellAmt / price), matching
+        // FixedPointLib.divIntegerByFixedPoint.
+        uint256 newBuyAmt = (newSellAmt << 63) / priceRepr;
```
duncancmt: If we need to handle overflow, left shift should be avoided. If we _don't_ need to handle overflow `/` should become `unsafeDiv`

src/core/Renegade.sol
```diff
@@ -5,110 +5,145 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
...
+
+        // Check newBuyAmt
+        if (newBuyAmt < minBuyAmt) revertTooMuchSlippage(buyToken, minBuyAmt, newBuyAmt);
+        require(newBuyAmt >= minInternalPartyAmountIn, "Renegade: newBuyAmt < minInternalPartyAmountIn");
+        require(newBuyAmt <= maxInternalPartyAmountIn, "Renegade: newBuyAmt > maxInternalPartyAmountIn");
```
duncancmt: Presumably Renegade checks this internally? If it doesn't, please replace this with the pattern:
```
if (!condition) {
    revert MyCustomErrorSelector(...);
}
```

or, even better, throw the error in assembly. This is for contract size. If this isn't already documented as a stylistic dictum in `AGENTS.md`, please add it.

src/core/Renegade.sol
```diff
@@ -5,110 +5,145 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
...
+
+        require(target == _renegadeGasSponsorV2(), "Renegade: bad target");
+
+        // Snapshot the buyToken balance before the call so we can measure the
+        // actual amount transferred after.
+        uint256 buyTokenBalanceBefore = buyToken.fastBalanceOf(address(this));
```
duncancmt: This is an antipattern for slippage checking https://kebabsec.xyz/posts/critical_vulnerability_in_uniswapx/

src/core/Renegade.sol
```diff
@@ -5,110 +5,145 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
...
+            // Restore the length word clobbered with the selector. Required to
+            // (a) keep `data` a valid `bytes memory` for any subsequent reads, and
+            // (b) honor the memory-safe assembly contract, which forbids leaving
+            // compiler-managed slots in a corrupted state on exit.
             mstore(data, len)
         }
+
+        // Check the balance delta.
+        buyAmt = buyToken.fastBalanceOf(address(this)) - buyTokenBalanceBefore;
+        if (buyAmt < minBuyAmt) revertTooMuchSlippage(buyToken, minBuyAmt, buyAmt);
```
duncancmt: Shouldn't this be handled by `_checkSlippageAndTransfer`? Also is custody optimization not possible with RenegadeV2?

duncancmt (commented):
General style in this repo is that constant arguments go on the left of commutative operations in assembly. E.g. `add(0x40, data)` ⭕ -- `add(data, 0x40)` ❌ 

If this isn't already in `AGENTS.md`, it should be

src/core/Renegade.sol
```diff
@@ -3,112 +3,104 @@ pragma solidity ^0.8.25;
...
+     *     0x80   address   internalPartyOutputToken    | BoundedMatchResult
+     *     0xa0   uint256   price.repr (FixedPoint)     | (inlined static struct,
+     *     0xc0   uint256   minInternalPartyAmountIn    |  six 32-byte words)
+     *     0xe0   uint256   maxInternalPartyAmountIn    |
+     *     0x100  uint256   blockDeadline              /
+     *     0x120  uint256   <offset>  SettlementBundle: (bool isFirstFill, uint8 bundleType, bytes data)
+     *     0x140  uint256   <offset>  GasSponsorOptions: (address refundAddress, bool refundNativeEth, uint256 refundAmount, uint256 nonce, bytes signature)
+     *   )
+     */
+    uint32 internal constant RENEGADE_SELECTOR = uint32(
```
duncancmt: If you define this constant as a literal, then you'll be able to access it in assembly. Even though the current expression is compile-time constant, because it isn't literal, it can't be accessed from Yul. Also should probably be `private`

src/chains/Base/Common.sol
```diff
@@ -105,16 +109,17 @@ abstract contract BaseMixin is
...
-                abi.decode(data, (address, IERC20, uint256, IEulerSwap, bool, uint256));
+            /*
+            } else if (action == uint32(ISettlerActions.EULERSWAP.selector)) {
+                (address recipient, IERC20 sellToken, uint256 bps, IEulerSwap pool, bool zeroForOne, uint256 amountOutMin) =
+                    abi.decode(data, (address, IERC20, uint256, IEulerSwap, bool, uint256));
 
-            sellToEulerSwap(recipient, sellToken, bps, pool, zeroForOne, amountOutMin);
-        */
+                sellToEulerSwap(recipient, sellToken, bps, pool, zeroForOne, amountOutMin);
+            */
```
duncancmt: please fix bad formatting

src/core/Renegade.sol
```diff
@@ -3,112 +3,92 @@ pragma solidity ^0.8.25;
...
+            buyToken := mload(add(0x60, data))
+            internalPartyOutputToken := mload(add(0x80, data))
+            priceRepr := mload(add(0xa0, data))
+            minInternalPartyAmountIn := mload(add(0xc0, data))
+            maxInternalPartyAmountIn := mload(add(0xe0, data))
         }
-    }
+        if ((recipient != address(0)) && (recipient != address(this))) revertInvalidRenegadeData();
+        if (sellToken != internalPartyOutputToken) revertInvalidRenegadeData();
+        if (priceRepr == 0) revertInvalidRenegadeData();
```
duncancmt: Join these together with `FastLogic.or` into a single branch to save bytecode size.

src/core/Renegade.sol
```diff
@@ -3,112 +3,92 @@ pragma solidity ^0.8.25;
...
-        } else {
-            newQuoteAmount = newSellAmount;
-            unchecked {
-                newBaseAmount = (originalBaseAmount * newQuoteAmount).unsafeDiv(originalQuoteAmount);
-            }
-            buyAmount = newBaseAmount;
-        }
+        if (newBuyAmt < minBuyAmt) revertTooMuchSlippage(buyToken, minBuyAmt, newBuyAmt);
+        if (newBuyAmt < minInternalPartyAmountIn) revertInvalidRenegadeData();
+        if (newBuyAmt > maxInternalPartyAmountIn) revertInvalidRenegadeData();
```
duncancmt: combine these with the above checks as well, presuming that these checks are meaningful and not redundant, which I suspect they are

src/core/Renegade.sol
```diff
@@ -3,112 +3,92 @@ pragma solidity ^0.8.25;
...
 
-        if (buyAmount < minBuyAmount) {
-            revertTooMuchSlippage(_extractBuyToken(data, baseForQuote), minBuyAmount, buyAmount);
-        }
+        address target = _renegadeGasSponsorV2();
+        uint256 buyTokenBalanceBefore = buyToken.fastBalanceOf(address(this));
```
duncancmt: This is still the forbidden way of checking slippage. Please fix.

src/core/Renegade.sol
```diff
@@ -3,112 +3,92 @@ pragma solidity ^0.8.25;
...
-            mstore(add(0x40, data), newBaseAmount)
+            // Override sellTokenAmt in data (at position 0x20) with newSellAmt.
+            mstore(add(0x20, data), newSellAmt)
 
+            // Stash the length and overwrite its slot with the selector; calldata
+            // starts at data + 0x1c so the call sees [selector | payload].
             let len := mload(data)
-            // temporarily clobber `data` size memory area
-            mstore(data, selector)
-            // Allowed selectors don't clash with any relevant function of restricted targets so we can skip checking `target`
```
duncancmt, line 104: Please restore this comment

src/core/Renegade.sol
```diff
@@ -3,112 +3,92 @@ pragma solidity ^0.8.25;
...
                 revert(ptr, returndatasize())
             }
-            // restore clobbered memory
+
+            // Restore the clobbered length word before leaving memory-safe assembly.
             mstore(data, len)
         }
+
+        buyAmt = buyToken.fastBalanceOf(address(this)) - buyTokenBalanceBefore;
+        if (buyAmt < minBuyAmt) revertTooMuchSlippage(buyToken, minBuyAmt, buyAmt);
```
duncancmt: Forbidden

src/ISettlerActions.sol
```diff
@@ -7,7 +7,6 @@ interface ISettlerActions {
     /// VIP actions should always start with `recipient` address and the `permit` from the taker
     /// followed by all the other parameters to ensure compatibility with `executeWithPermit` entrypoint.
     /// `minBuyAmount`/`amountOutMin` should always be the last parameter.
-
```
duncancmt, line 10: Please fix bad formatting

src/core/Renegade.sol
```diff
@@ -3,82 +3,49 @@ pragma solidity ^0.8.25;
...
         if (sellToken != internalPartyOutputToken) revertInvalidRenegadeData();
-        if (priceRepr == 0) revertInvalidRenegadeData();
 
         uint256 newSellAmt = sellToken.fastBalanceOf(address(this));
-
-        // newBuyAmt = floor(newSellAmt / price), matching
-        // FixedPointLib.divIntegerByFixedPoint.
-        uint256 newBuyAmt = (newSellAmt << 63).unsafeDiv(priceRepr);
-
-        if (newBuyAmt < minBuyAmt) revertTooMuchSlippage(buyToken, minBuyAmt, newBuyAmt);
```
duncancmt: Why was this removed? Does Renegade not support custody optimization?

src/core/Renegade.sol
```diff
@@ -3,82 +3,49 @@ pragma solidity ^0.8.25;
...
-        if (newBuyAmt < minInternalPartyAmountIn) revertInvalidRenegadeData();
-        if (newBuyAmt > maxInternalPartyAmountIn) revertInvalidRenegadeData();
```
duncancmt: just an idle thought here: should we clamp? if there's a maximum notional, if the swap encounters slippage, we should clamp to avoid a revert, no?

src/core/Renegade.sol
```diff
@@ -3,82 +3,49 @@ pragma solidity ^0.8.25;
...
-     *     0xe0   uint256   maxInternalPartyAmountIn    |
-     *     0x100  uint256   blockDeadline              /
-     *     0x120  uint256   <offset>  SettlementBundle: (bool isFirstFill, uint8 bundleType, bytes data)
-     *     0x140  uint256   <offset>  GasSponsorOptions: (address refundAddress, bool refundNativeEth, uint256 refundAmount, uint256 nonce, bytes signature)
-     *   )
-     */
+    // `data` is the args to `sponsorExternalMatch` minus the 4-byte selector; its payload begins at
+    // `data + 0x20`. The head is static (`BoundedMatchResult` is inlined), so we index fixed offsets:
+    // 0x20 externalPartyAmountIn, 0x40 recipient, 0x80 internalPartyOutputToken.
     uint32 private constant RENEGADE_SELECTOR = 0x54ea46d4;
```
duncancmt, line 14: it doesn't necessarily need to be here, but somewhere there should be a more readable signature from which this selector was computed

src/core/Renegade.sol
```diff
@@ -87,8 +54,5 @@ abstract contract Renegade is SettlerSwapAbstract {
             // Restore the clobbered length word before leaving memory-safe assembly.
             mstore(data, len)
         }
-
-        buyAmt = buyToken.fastBalanceOf(address(this)) - buyTokenBalanceBefore;
-        if (buyAmt < minBuyAmt) revertTooMuchSlippage(buyToken, minBuyAmt, buyAmt);
```
duncancmt: 👍

duncancmt (commented):
This is looking pretty good. I think I still don't all the way understand how we're expecting Renegade to interact with slippage checking and custody optimization. The previous iteration seemed like it was amenable to custody optimization (and consequently slippage checking), but the current iteration removes that functionality entirely. Is this because we don't know whether we'll be applying that optimization at the time we request the firm quote from Renegade and the recipient is cryptographically bound to the price?

Even if custody optimization isn't possible, it's still desirable to be able to slippage-check the Renegade leg. Per-leg slippage checking is an infrequent, but still-used idiom.

Lastly, if we're going to be destructuring/interpreting many of the fields of the arguments to `sponsorExternalMatch`, it would make sense for those fields to be arguments of the action rather than taking it as opaque bytes and implicitly extracting them. It's less error-prone and (sometimes) more gas efficient. If we're not reading any of the `BoundedMatchResult` fields, then that can still be opaque.

Feel free to DM me if you want to iterate faster.

src/core/Renegade.sol
```diff
@@ -3,112 +3,89 @@ pragma solidity ^0.8.25;
...
 
-        // word 0: quoteAmount, word 1: baseAmount
-        uint256 originalQuoteAmount;
-        uint256 originalBaseAmount;
-        assembly ("memory-safe") {
-            originalQuoteAmount := mload(add(0x20, data))
-            originalBaseAmount := mload(add(0x40, data))
-        }
+        uint256 sellAmt = sellToken.fastBalanceOf(address(this));
+        if (sellAmt > maxSellAmount) sellAmt = maxSellAmount;
```
duncancmt, line 34: You are close to Solver than I am. Is the lack of a `bps` parameter here desirable?

src/core/Renegade.sol
```diff
@@ -3,112 +3,89 @@ pragma solidity ^0.8.25;
...
-        }
 
-        if (buyAmount < minBuyAmount) {
-            revertTooMuchSlippage(_extractBuyToken(data, baseForQuote), minBuyAmount, buyAmount);
-        }
+            // Patch caller-malleable args and bind typed fields into proof-authenticated data.
+            mstore(add(0x20, data), sellAmt)
+            mstore(add(0x40, data), recipient)
+            mstore(add(0x60, data), buyToken)
+            mstore(add(0x80, data), sellToken)
```
duncancmt: This is a nit and I'm not sure there's a satisfying fix, but carrying 4 extra words of `data` just to obliterate them at settlement time seems _very_ wasteful

src/core/Renegade.sol
```diff
@@ -3,112 +3,89 @@ pragma solidity ^0.8.25;
...
+
+            if lt(returndatasize(), 0x20) {
+                mstore(0x00, 0xaa81f37c) // selector for `InvalidRenegadeData()`
+                revert(0x1c, 0x04)
+            }
+            buyAmt := mload(0x00)
+        }
+
+        if (!refundNativeEth || buyToken == ETH_ADDRESS) {
+            if (buyAmt > maxRefundAmount) buyAmt -= maxRefundAmount;
```
duncancmt: `unchecked`

duncancmt (approved):
This is very nice. 🔥

src/core/Renegade.sol
jparklev: done f1af5bb1b9ea008382269afa243def29e079fc86

src/core/Renegade.sol
jparklev: Should be better in d5cfecc0e. Action now omits the selector and the four duplicated words, and we reconstruct the sponsor call in a fresh buffer. Can update encoding on the Solver side

src/core/Renegade.sol
jparklev: Yes, I think so. We treat it like Bebop (quote-capped by `maxSellAmount`) since it’s similarly firm-ish and solver always runs it first (+ adding bps would add a second way for amt out to fall outside of renegade's proof bounds)

src/core/Renegade.sol
```diff
@@ -3,112 +3,90 @@ pragma solidity ^0.8.25;
...
+                revert(0x1c, 0x04)
+            }
+            buyAmt := mload(0x00)
+        }
+
+        if (!refundNativeEth || buyToken == ETH_ADDRESS) {
+            unchecked {
+                if (buyAmt > maxRefundAmount) buyAmt -= maxRefundAmount;
+                else buyAmt = 0;
+            }
```
duncancmt: I should've suggested this the first time around 🤦 

https://github.com/0xProject/0x-settler/blob/master/src/utils/UnsafeMath.sol#L145-L149

duncancmt (approved):

duncancmt (approved):

duncancmt (approved):

duncancmt (approved):

duncancmt (approved):

jparklev:
sigh. ci was failing from an incorrect import. fixed now. apologies for the noise

PR #576, opened 2026-06-02
Dry allowanceHolder use in test

e1Ru1o:


test/unit/BridgeSettler.t.sol
```diff
@@ -31,10 +31,10 @@ contract BridgeDummy {
 }
 
 contract BridgeSettlerTestBase is Test {
+    IAllowanceHolder allowanceHolder = ALLOWANCE_HOLDER;
```
duncancmt: Can we make this a constant to make the gas profiling more accurate?

test/integration/SettlerBasePairTest.t.sol
```diff
@@ -11,8 +11,9 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
 import {LibBytes} from "../utils/LibBytes.sol";
 import {SafeTransferLib} from "src/vendor/SafeTransferLib.sol";
 
-import {IAllowanceHolder} from "src/allowanceholder/IAllowanceHolder.sol";
-import {MainnetSettler as Settler} from "src/chains/Mainnet/TakerSubmitted.sol";
+import {IAllowanceHolder, ALLOWANCE_HOLDER} from "src/allowanceholder/IAllowanceHolder.sol";
+import {MainnetSettler} from "src/chains/Mainnet/TakerSubmitted.sol";
+import {Settler} from "src/Settler.sol";
 
 contract Shim {
```
duncancmt: Is this contract still in use anywhere? We should be able to use `vm.getChainId()` everywhere now

test/integration/SettlerBasePairTest.t.sol
duncancmt: Ditto previous comment. Can we make this constant/immutable to improve gas profiling?

test/integration/SettlerBasePairTest.t.sol
e1Ru1o: Yep, that is used in some files, I can remove it.

e1Ru1o:
Unit tests are passing as well, CI is failing because of the SafeGuard steps in master that are not in this branch

PR #585, opened 2026-06-14
Add formally verified natural logarithm

duncancmt:


test/0.8.25/formal-model/LnModel.t.sol
e1Ru1o, line 1: Consider also adding a diff test for the python implementation in `check_ln_counterexample.py`

PR #592, opened 2026-06-30
Fix `metaTx` malleability in CrossChainReceiverFactory in response to Immunefi bug 78645

duncancmt:


e1Ru1o (approved):

jparklev (approved):

jparklev:
One note. The previous code, I think, allowed a signed call to verify when:
- `Call.target` was the literal proxy, and
- `metaTx` was called with `owner_ == 0`

And now it won't (hashes as the sentinel instead). Which, I think is correct.

PR #590, opened 2026-06-26
Add Safe{Wallet} v1.4.1 config to `chain_config.json`

duncancmt:

PR #597, opened 2026-07-10
Add SafeGuard deployment script

e1Ru1o:


sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,324 @@
...
+        ZeroExSettlerDeployerSafeGuardOnePointThree
+        ZeroExSettlerDeployerSafeGuardOnePointFourPointOne
+    )
+fi
+
+function predict_create2 {
+    declare _predict_out
+    _predict_out="$(cast keccak "$(cast concat-hex 0xff "$1" "$(cast hash-zero)" "$2")")"
+    cast to-check-sum-address "0x${_predict_out:26:40}"
+}
```
duncancmt, line 188: Use `cast compute-address`

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,324 @@
...
+    [ZeroExSettlerDeployerSafeGuardOnePointFourPointOne]=0x3555bd3ee95b1c6605c602740d71efaf200068e0395ccd701ac82ab8e42307bd
+    [ZeroExSettlerDeployerSafeGuardOnePointThreeEraVm]=0x0100080f935a1a562e892e1e71d9a0ca8cd349d19a413e0b7e7172c5e8c83ed1
+    [ZeroExSettlerDeployerSafeGuardOnePointFourPointOneEraVm]=0x010006c19437ff25b448f038f7ea0a4c910e0ae9cd8e55f2d199b7916b72eb1e
+)
+if [[ $era_vm != [Ff]alse ]] ; then
+    candidate_factories=(0xaECDbB0a3B1C6D1Fe1755866e330D82eC81fD4FD)
+    candidate_variants=(
+        ZeroExSettlerDeployerSafeGuardOnePointThreeEraVm
+        ZeroExSettlerDeployerSafeGuardOnePointFourPointOneEraVm
+    )
```
duncancmt: seems awkward that you're repeating the keys of `singleton_inithash` here. can you DRY this by reshaping these mappings?

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,324 @@
...
+safe="$(get_config governance.upgradeSafe)"
+declare -r safe
+
+if [[ ${safe:-null} == [nN][uU][lL][lL] ]] || [[ -z ${safe:-} ]] ; then
+    echo 'governance.upgradeSafe is missing from chain_config.json for chain "'"$chain_name"'"' >&2
+    exit 1
+fi
+
+declare onchain_singleton
+onchain_singleton="$(cast call --rpc-url "$rpc_url" "$safe" 'masterCopy()(address)')"
```
duncancmt, line 162: needs a check against the codehash before you can trust `masterCopy()`

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,324 @@
...
+# The Guard MUST compile with these exact settings: 
+export FOUNDRY_EVM_VERSION=london
+export FOUNDRY_OPTIMIZER_RUNS=200
+
+declare constructor_args
+constructor_args="$(cast abi-encode 'constructor(address)' "$safe")"
+declare -r constructor_args
+
+declare predicted
+if [[ $era_vm = [Ff]alse ]] ; then
```
duncancmt: I don't have a strong opinion, but it's worth giving a moment of thought here what you want the behavior to be in the case of a malformed `$era_vm`. Do you want to default to EraVM or EVM?

duncancmt (commented):
This should probably build on top of #590 . Also some of the changes from https://github.com/0xProject/0x-settler/tree/dcmt/safe-v1.4.1-installation may be useful. Feel free to merge them or to cherry-pick them, as you see fit

Otherwise, looks pretty good!

sh/deploy_safeguard.sh
e1Ru1o: I want to default to EVM so just modified it

e1Ru1o:
Atm I am not using anything from the `safe` field in the script even after resolving your comments, so if we don't change the behavior to be more like your branch `safe-v1.4.1-installation` branch, I don't think I need to rebase #590. I see though that installation branch has some good features and does simulation to check that the Guard really works. I am fine using your installation branch, they handle in different ways forge versions but I am fine with any of the two.

As your is more feature extensive, consider opening a PR, I can close this one and do a follow up PR on top of yours with any changes I would recommend.

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+declare -r project_root
+cd "$project_root"
+
+. "$project_root"/sh/common.sh
+
+declare safe
+safe="$(get_config governance.upgradeSafe)"
+declare -r safe
+
+if [[ ${safe:-null} == [nN][uU][lL][lL] ]] || [[ -z ${safe:-} ]] ; then
```
duncancmt: This is unnecessarily defensive. Consider using a more intelligent LLM model

```suggestion
if [[ ${safe:-null} == [nN][uU][lL][lL] ]] ; then
```

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+else
+    safe_codehash="$(cast keccak "$(cast code --rpc-url "$rpc_url" "$safe")")"
+fi
+declare -r safe_codehash
+case "$safe_codehash" in
+    0xaea7d4252f6245f301e540cfbee27d3a88de543af8e49c5c62405d5499fab7e5|\
+    0xb89c1b3bdf2cf8827818646bce9a8f6e372885f8c55e5c07acbd307cb133b000|\
+    0xd7d408ebcd99b2b70be43e20253d6d92a8ea8fab29bd3be7f55b10032331fb4c|\
+    0x0100004124426fb9ebb25e27d670c068e52f9ba631bd383279a188be47e3f86d|\
+    0x0100003b6cfa15bd7d1cae1c9c022074524d7785d34859ad0576d8fab4305d4f) ;;
```
duncancmt: Separate these by `$era_vm` cases so that we don't accept EVM on EraVM and vice versa

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+    *)
+        echo 'Upgrade Safe ('"$safe"') is not a recognized Safe proxy (codehash '"$safe_codehash"')' >&2
+        exit 1
+        ;;
+esac
+
+declare onchain_singleton
+onchain_singleton="$(cast call --rpc-url "$rpc_url" "$safe" 'masterCopy()(address)')"
+onchain_singleton="$(cast to-check-sum-address "$onchain_singleton")"
```
duncancmt: Unnecessary defensiveness

```suggestion
```

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+
+function predict_create2 {
+    declare _predict_out
+    _predict_out="$(cast compute-address "$1" --salt "$(cast hash-zero)" --init-code-hash "$2")"
+    echo "${_predict_out##* }"
+}
+
+function predict_create2_era_vm {
+    declare _predict_out
+    _predict_out="$(cast keccak "$(cast concat-hex "$(cast keccak zksyncCreate2)" "$(cast to-uint256 "$1")" "$(cast hash-zero)" "$2" "$3")")"
```
duncancmt, line 196: I don't think potentially-failing expansions inside output-capturing work the way that you want to. Double check that passing short arguments to this function fails and bubbles like you expect

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+    cast to-check-sum-address "0x${_predict_out:26:40}"
+}
+
+function predict_singleton {
+    if [[ $era_vm = [Tt]rue ]] ; then
+        predict_create2_era_vm "$1" "$2" "$(cast keccak '')"
```
duncancmt: style

```suggestion
        predict_create2_era_vm "$1" "$2" "$(cast keccak 0x)"
```

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+if [[ $era_vm = [Tt]rue ]] ; then
+    # EraVM needs zkSync artifacts so we switch to the zksync aware foundry version
+    foundryup-zksync -u foundry-zksync-v0.1.9 || true
+    if [[ $(forge --version) != *14afc70e251c89b7e2af6e6ac02e9ac6f095b5cc* ]] ; then
+        echo 'Wrong foundry version installed' >&2
+        echo 'Run `foundryup-zksync -i foundry-zksync-v0.1.9`' >&2
+        exit 1
+    fi
+    # foundry-zksync errors on an empty profile, so, we set default if empty
+    export FOUNDRY_PROFILE="${FOUNDRY_PROFILE:-default}"
```
duncancmt: This is really good functionality. Can you move this into `common.sh` and update `foundry.toml` so that the patch is unnecessary?

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+
+    forge clean
+    forge build --zksync --zk-compile src/deployer/SafeGuard.sol
```
duncancmt, line 260: Can this pattern of passing `--zksync` and `--zk-compile` be reified? Alternatively, is this the only site where we deliberately build a EraVm contract instead of EVM?

If it's reusable, it would be good to move this into `common.sh` and refactor all the sites where we invoke `forge build` (or similar) to pass around these flags, as necessary.

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+    declare art="$project_root/zkout/SafeGuard.sol/$guard_contract.json"
+    declare bytecode_hash guard_bytecode
+    bytecode_hash="0x$(jq -Mr '.hash' "$art")"
+    guard_bytecode="0x$(jq -Mr '.bytecode.object' "$art")"
```
duncancmt, line 264: Can this be done with `forge inspect` instead of parsing the artifact with `jq`?

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+    echo "${_predict_out##* }"
+}
+
+function predict_create2_era_vm {
+    declare _predict_out
+    _predict_out="$(cast keccak "$(cast concat-hex "$(cast keccak zksyncCreate2)" "$(cast to-uint256 "$1")" "$(cast hash-zero)" "$2" "$3")")"
+    cast to-check-sum-address "0x${_predict_out:26:40}"
+}
+
+function predict_singleton {
```
duncancmt: This function isn't deduplicating anything. Inline its functionality at its only callsite

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+    declare -r guard_bytecode initcode
+    predicted="$(predict_create2 "$factory" "$(cast keccak "$initcode")")"
+fi
+declare -r predicted
+
+echo 'SafeGuard variant : '"$guard_contract" >&2
+echo 'Protected Safe    : '"$safe" >&2
+echo 'Predicted address : '"$predicted" >&2
+
+if [[ "$(cast code --rpc-url "$rpc_url" "$predicted")" != '0x' ]] ; then
```
duncancmt: Quoting on the left is unnecessarily defensive inside `[[`
```suggestion
if [[ $(cast code --rpc-url "$rpc_url" "$predicted") != '0x' ]] ; then
```

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,338 @@
...
+
+cast "${maybe_broadcast[@]}" --from "$signer" --rpc-url "$submit_rpc" --gas-price $gas_price --gas-limit $gas_limit "${extra_flags[@]}" "${zk_tx_flags[@]}" "${deploy_args[@]}"
+
+if [[ ${BROADCAST-no} = [Yy]es ]] ; then
+    sleep 60
+
+    if [[ "$(cast code --rpc-url "$rpc_url" "$predicted")" == '0x' ]] ; then
+        echo 'Deployment did not produce code at the predicted address '"$predicted" >&2
+        exit 1
+    fi
```
duncancmt: Unnecessarily defensive. We should never hit this, and if we did `verify_contract` would fail anyways

sh/deploy_safeguard.sh
e1Ru1o: foundry-zksync forge inspect is broken, it seems that it doesn't resolve remappings correctly when using `--zksync` flag, if not used it gives the one from EVM output

sh/deploy_safeguard.sh
e1Ru1o: yes, this is the only one, and it is just because we need `.hash` and eraVM bytecode output in the artifact. I don't think we need to reuse it at the moment.

sh/deploy_safeguard.sh
e1Ru1o: About the patch, yes. As for moving this version switch to `common.sh` I have been thinking about it since I saw the branch that you shared but, using this is very specific for the create2 deployment that we need to do here, if it wasn't for that we wouldn't need any specific zksync feature. Maybe switching to foundry-zksync is harmless but I prefer to only do it if it is strictly necessary as this toolchain is less trusted.
At the moment I would keep it just here

sh/deploy_safeguard.sh
e1Ru1o: added size checks on the inputs, to avoid errors here

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,336 @@
...
+declare -r safe
+
+if [[ ${safe:-null} == [nN][uU][lL][lL] ]] ; then
+    echo 'governance.upgradeSafe is missing from chain_config.json for chain "'"$chain_name"'"' >&2
+    exit 1
+fi
+
+declare safe_codehash
+if [[ $era_vm = [Tt]rue ]] ; then
+    safe_codehash="$(cast call --rpc-url "$rpc_url" 0x0000000000000000000000000000000000008002 'getCodeHash(uint256)(bytes32)' "$safe")"
```
duncancmt, line 138: This warrants a comment referencing the relevant EraVM docs

sh/deploy_safeguard.sh
```diff
@@ -0,0 +1,336 @@
...
+    singleton_inithash=(
+        [ZeroExSettlerDeployerSafeGuardOnePointThree]=0x49f30800a6ac5996a48b80c47ff20f19f8728812498a2a7fe75a14864fab6438
+        [ZeroExSettlerDeployerSafeGuardOnePointFourPointOne]=0x3555bd3ee95b1c6605c602740d71efaf200068e0395ccd701ac82ab8e42307bd
+    )
+fi
+
+function predict_create2 {
+    declare _predict_out
+    _predict_out="$(cast compute-address "$1" --salt "$(cast hash-zero)" --init-code-hash "$2")"
+    echo "${_predict_out##* }"
```
duncancmt: Is there a flag we can pass to suppress extraneous output rather than slicing it?

duncancmt (approved):

sh/deploy_safeguard.sh
e1Ru1o: There is none, it is sliced everywhere else it is used in the scripts. But, `cast create2` only outputs the address, so, changed to that.

duncancmt: I did not know `cast create2` could do that. I thought it was just for mining vanity addresses

duncancmt (approved):

PR #616, opened 2026-07-28
Add Bailsec AllowanceHolder Formal Verification report and files

e1Ru1o:


jparklev (approved):
I think `AllowanceHolderHarness.sol` doesn’t compile atm (references `_rejectIfERC20Inner()` eg, generally missing source for some things). Is that ok?

jparklev (approved):

PR #617, opened 2026-07-30
Migrate proportional action arguments from bps to ppm

e1Ru1o:
## Summary

**Breaking change.** Every action argument that expresses a proportion of Settler's token balance is re-denominated from basis points (`BASIS = 10_000`) to parts-per-million (`BASIS = 1_000_000`), and renamed `bps` → `ppm` (`maxBps` → `maxPpm` in `POSITIVE_SLIPPAGE`). This makes sub-bp proportions expressible (e.g. a 30 ppm fee) and stops small fee proportions from truncating to zero on large-notional trades.

Action selectors are unchanged (parameter names are not part of the selector), so **the unit is a property of the Settler deployment**: encoders must key bps vs ppm on the Settler address they target.

## Details

- `BASIS` becomes `1_000_000` in `SettlerAbstract`, `Encoder`, and `Decoder` (the constructor cross-check asserts are unchanged and still hold).
- The packed proportion field widens from 2 bytes to 3 bytes in two places:
  - the non-VIP flash-accounting header (`Encoder.encode` / `Decoder.initialize` in `FlashAccountingCommon.sol`); VIP headers carry no proportion and are untouched;
  - the leading field of every fill in `UNISWAPV4`, `BALANCERV3`, `PANCAKE_INFINITY`, `EKUBO`, and `EKUBOV3`. `_HOP_DATA_LENGTH` grows by 1 per venue (52→53, 3→4, 82→83, 47→48, 47→48).
- Flag bits that ride on the fill proportion move to the new MSBs: BalancerV3 wrap/unwrap bits 15/14 → 23/22 (masks `0xc00000`/`0x400000`/`0x3fffff`), Ekubo forwarding-extension bit 15 → 23 (`0x800000`/`0x7fffff`).
- The Permit2 balance-proportional sell amount sentinel (`~amount < BASIS`) and the RFQ maker-permit asserts widen automatically with the constant.
- The `[Unreleased]` CHANGELOG entry documents the full encoding change for integrators; `sh/initial_description_*.md` are updated so future chain bring-ups register `ppm` wording (existing chains' feature descriptions are immutable one-shot IPFS metadata and are unaffected).

### Deliberately unchanged

Pool fees carried in the upper 16 bits of `swapInfo` (UniswapV2, Velodrome, VelodromeAlt) stay denominated in bps against a hard-coded `10_000`: they describe the external pool's own fee convention, and 1e6 does not fit in that 16-bit field. Likewise EulerSwap's 1e4 LTV scale and MakerPSM's `GEM_basis`.

## Testing

- Unit tests: 189/189 pass via the CI command, including the UniswapV4 suite exercising the 3-byte fill encoding (single/multihop/multiplex/diamond/VIP) against a real PoolManager.
- All unit and integration tests migrated: proportional values ×100, fill encoders `uint16`/`bytes2` → `uint24`/`bytes3`, flag constructions moved to bits 23/22.
- `FOUNDRY_PROFILE=integration` compiles; fork tests + gas comparison to be confirmed by CI.
- Per-chain `forge build --sizes` passes for every chain in `chain_config.json`.

## Gas Optimization

No hot-path changes; the arithmetic is identical (`* ppm / BASIS`). Calldata grows 1 byte per fill plus 1 byte per flash-accounting action. Contract size deltas vs `master` (95184a233) on the tightest chain:

| Contract | master | this PR | Δ | headroom |
|---|---|---|---|---|
| MainnetSettler | 23,619 | 23,625 | +6 B | 951 B |
| MainnetSettlerIntent | 22,487 | 22,491 | +4 B | 2,085 B |
| MainnetSettlerMetaTxn | 21,661 | 21,660 | −1 B | 2,916 B |
| MainnetBridgeSettler | 6,340 | 6,345 | +5 B | 18,231 B |

The growth is the `PUSH2`→`PUSH3` widening of the `BASIS` literal and flag masks.


test/integration/SettlerPairTest.t.sol
```diff
@@ -73,7 +73,7 @@ abstract contract SettlerPairTest is SettlerBasePairTest {
                 ISettlerActions.BASIC,
                 (
                     address(fromToken()),
-                    10_000,
+                    1_000_000,
```
jparklev, line 76: Should we add some sub-bp cases here to test? (eg the 30 ppm fee mentioned by Phil)

src/core/FlashAccountingCommon.sol
```diff
@@ -481,7 +481,7 @@ library Decoder {
     using NotesLib for NotesLib.Note;
     using NotesLib for NotesLib.Note[];
 
-    uint256 internal constant BASIS = 10_000;
+    uint256 internal constant BASIS = 1_000_000;
```
jparklev, line 484: thoughts on one shared like `PPM_DENOM` constant we import? (should be runtime bytecode neutral)

(in theory we also then wouldn't have to do like `assert(BASIS == Encoder.BASIS)` in BalancerV3 and pancake constructors)

src/core/FlashAccountingCommon.sol
```diff
@@ -653,38 +653,38 @@ library Decoder {
...
 
-            uint16 bps;
+            uint24 ppm;
             assembly ("memory-safe") {
                 // `data` hasn't been advanced from decoding `sellToken` above. so we have to
-                // implicitly advance it by 20 bytes to decode `bps` then advance by 22 bytes
+                // implicitly advance it by 20 bytes to decode `ppm` then advance by 23 bytes
 
-                bps := shr(0x50, calldataload(data.offset))
+                ppm := shr(0x48, calldataload(data.offset))
```
jparklev, line 661: nit: consider `uint256 ppm; + ppm := shr(0xe8, calldataload(add(0x14, data.offset)))`.

It’s a couple bytes smaller, and then we don't need the “implicitly advance by 20 bytes” comment

(similar in erc20 branch below. could punt to a future cleanup)

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ### Breaking changes
 
+* All proportional-amount action arguments are now denominated in parts-per-million (`ppm`; denominator 1_000_000) instead of basis points (`bps`; denominator 10_000)
+  * Affects the third-ish argument of `UNISWAPV3`, `UNISWAPV2`, `BASIC`, `VELODROME`, `MAKERPSM`, `DODOV1`, `DODOV2`, `MAVERICKV2`, `EULERSWAP`, `HANJI`, `UNISWAPV4`, `BALANCERV3`, `PANCAKE_INFINITY`, `EKUBO`, and `EKUBOV3`; the `maxBps` (now `maxPpm`) argument of `POSITIVE_SLIPPAGE`; and BridgeSettler's `BASIC`
```
jparklev, line 6: "third-ish" reads as kind of odd here to me. Maybe just named like "the proportion-of-balance argument"?

jparklev (commented):
Nice one.

Should we mention in the README like, "the first ppm deployment on each chain is nonce N (per-chain table), earlier nonces are bps”? Not 100% whats best for integrators, but seems like a pretty large behavoiral change we wouldn't want anyone to opt into accidentally

test/integration/SettlerPairTest.t.sol
e1Ru1o: we might, but: 
- ideally I would keep the proportions of all tests to not break them.
- I don't think it brings that much value as tests should already test different proportions

src/core/FlashAccountingCommon.sol
e1Ru1o: yeah, at some point we should get to have a `Constants.sol` for things like this and native for example

src/core/FlashAccountingCommon.sol
e1Ru1o: that is a good catch

CHANGELOG.md
e1Ru1o: changed it to `bps` to be explicit about the argument affected instead of its position

src/core/FlashAccountingCommon.sol
duncancmt: Seems reasonable to me. Maybe add it to a `Constants.sol` file where we also put the ERC7528 constant?

duncancmt: Lol. I am slow.

src/core/FlashAccountingCommon.sol
duncancmt: I believe the original implementation that created a dirty `bps` was more gas/stack efficient because it avoided an extra `calldataload`. But this is relatively old code that was written for an older `solc` that might be dumber than the current one.

It's worth testing.

duncancmt (approved):

src/core/FlashAccountingCommon.sol
e1Ru1o: Created the file. Please check if you are ok with the way it is used in `SettlerAbstract.sol`, I decided to keep it like that to not have a bunch of other files having to import the constants but that is a simple change that I can do as well.

e1Ru1o:
> Should we mention in the README ...

Do you think the CHANGELOG.md is not enough to cover that? I don't know how would it work for integrators, but in API terms if they set the bps on their own I would expect the API team to accept the parameter and adjust it to ppm, and, ofc, advice them to start using ppm instead.
On the other hand I don't know if we have integrators that are using the contracts on their own instead of provided by us (the API)

jparklev:
> > Should we mention in the README ...
> 
> Do you think the CHANGELOG.md is not enough to cover that? I don't know how would it work for integrators, but in API terms if they set the bps on their own I would expect the API team to accept the parameter and adjust it to ppm, and, ofc, advice them to start using ppm instead. On the other hand I don't know if we have integrators that are using the contracts on their own instead of provided by us (the API)

Yea, I think `CHANGELOG` is probably enough. Tbh I don't know what's best -- what you say makes sense. Mainly had the thought, as I went through, that, "anybody expecting bps is going to have a bad time". Probably worth checking with the api team, and maybe Doom to see if we know of anybody using the contracts directly. Also for some data use cases where ppl read the calldata it could be relevant

src/core/FlashAccountingCommon.sol
e1Ru1o: In terms of gas reverting to the old approach all integrations using FlashAccounting get +3 units of gas in the test snapshots, so seems that the new approach is better in gas too

src/core/FlashAccountingCommon.sol
e1Ru1o: I also did not changed the tests, so changes in the constants breaks them, specially changes related to ppm. Native can be used from constants in the test as it is not meant to change at all.

jparklev (approved):

src/SettlerAbstract.sol
```diff
@@ -5,6 +5,7 @@ import {IERC20} from "@forge-std/interfaces/IERC20.sol";
 import {ISettlerBase} from "./interfaces/ISettlerBase.sol";
 import {Permit2PaymentAbstract} from "./core/Permit2PaymentAbstract.sol";
 import {uint512} from "./utils/512Math.sol";
+import "./core/Constants.sol" as Constants;
```
duncancmt, line 8: I did not know you could use `as` for an entire file

duncancmt (approved):

src/SettlerAbstract.sol
e1Ru1o: We might be able to clean up some long imports if we start using that

PR #619, opened 2026-08-03
Deploy the SafeGuard on EraVM chains through Frame

e1Ru1o:
## What

`sh/deploy_safeguard.sh` gains a broadcast path for EraVM chains that works with the `frame` wallet type, and `cast estimate` no longer receives `--chain`.

## Why

Deploying a contract on an EraVM chain requires its bytecode to be published via the `factoryDeps` field, which exists only on zkSync's native EIP-712 transaction type (`0x71`). Wallets that submit through `eth_sendTransaction` (Frame included) only build standard Ethereum transaction types, so the factory deps are silently dropped and the deployment halts in the bootloader with an unknown code hash. Type-`0x71` transactions are authorized by an EIP-712 typed-data signature, which Frame does provide.

The foundry-zksync fork's `cast estimate` (active on the EraVM path) has no `--chain` flag. The flag is redundant on every chain: `sh/common.sh` already verifies the RPC's chain id against the config, and `cast send --chain` still guards the submission RPC.

## How

On `BROADCAST=yes` for an EraVM chain, the script:

1. rejects any wallet type other than `frame`;
2. builds the type-`0x71` transaction (deployment calldata to the CREATE2 factory, guard bytecode as the sole factory dep, fees from the existing gas estimation, protocol-default gas-per-pubdata of 50,000);
3. requests Frame's signature over the transaction's typed data via `eth_signTypedData_v4` against Frame's RPC, normalizing the recovery byte to {27, 28} as required by the bootloader's DefaultAccount;
4. RLP-serializes the transaction with the signature as `customSignature` using `cast to-rlp` (integers as JSON numbers, which encode canonically) and broadcasts it with `cast publish` to the chain's RPC.

Non-EraVM chains and the dry-run path are unchanged, and the flow lands in the existing verification step afterwards.

## Testing

- The serialized transaction and the signed EIP-712 digest are byte-identical to zksync-ethers' `serializeEip712` / `EIP712Signer.getSignedDigest` for identical inputs (including the SDK convention that the priority fee equals the max fee).
- This flow deployed the Abstract SafeGuard end to end: [`ZeroExSettlerDeployerSafeGuardOnePointFourPointOneEraVm` at `0x9Ee6c2b1F81eFC371e6fD5fb6A9D3578c36bb969`](https://abscan.org/address/0x9Ee6c2b1F81eFC371e6fD5fb6A9D3578c36bb969#code), verified on Abscan by the script's verification step.

## Gas Optimization

Shell tooling only; no contract changes and no gas impact.

### DIFF
--- a/sh/deploy_safeguard.sh
+++ b/sh/deploy_safeguard.sh
@@ -324,7 +324,114 @@ fi
 declare -r -a maybe_broadcast
 declare -r submit_rpc
 
-cast "${maybe_broadcast[@]}" --from "$signer" --rpc-url "$submit_rpc" --gas-price $gas_price --gas-limit $gas_limit "${extra_flags[@]}" "${zk_tx_flags[@]}" "${deploy_args[@]}"
+if [[ ${BROADCAST-no} = [Yy]es && $era_vm = [Tt]rue ]] ; then
+    # Frame only builds standard Ethereum transaction types, but publishing the guard
+    # bytecode rides exclusively on the factory-deps field of EraVM's native EIP-712
+    # transaction type (0x71). That type is authorized by an EIP-712 typed-data
+    # signature, which Frame does provide: collect Frame's signature over the
+    # transaction via eth_signTypedData_v4, then serialize the transaction ourselves
+    # and broadcast it directly to the chain's RPC.
+    if [[ $wallet_type != 'frame' ]] ; then
+        echo 'The EraVM deployment flow only supports the "frame" wallet type' >&2
+        exit 1
+    fi
+
+    declare -i nonce
+    nonce="$(cast nonce --rpc-url "$rpc_url" "$signer")"
+    declare -r -i nonce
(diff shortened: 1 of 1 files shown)


sh/deploy_safeguard.sh
```diff
@@ -324,7 +324,114 @@ fi
...
+    declare -i nonce
+    nonce="$(cast nonce --rpc-url "$rpc_url" "$signer")"
+    declare -r -i nonce
+    # protocol default gas-per-pubdata limit for type 0x71 transactions
+    declare -r -i gas_per_pubdata=0xc350
+
+    declare typed_data
+    typed_data="$(jq -Mcn \
+        --arg from "$(cast to-dec "$signer")" \
+        --arg to "$(cast to-dec "$factory")" \
```
duncancmt, line 348: This is very dumb 🙃 

Thank you for dealing with this

duncancmt (approved):

PR #621, opened 2026-08-07
Add OrvexCL PancakeInfinity fork to Robinhood

e1Ru1o:


jparklev (approved):
👍

src/core/PancakeInfinity.sol
```diff
@@ -506,33 +506,13 @@ abstract contract PancakeInfinity is SettlerSwapAbstract {
...
-                        zeroForOne
-                    );
-                } else {
-                    assembly ("memory-safe") {
-                        mstore(0x00, 0x0a9a7da6) // selector for `UnknownPoolManagerId(uint8)`
-                        mstore(0x20, and(0xff, poolManagerId))
-                        revert(0x1c, 0x24)
-                    }
-                }
+                _pancakeInfinitySettleDelta(
```
duncancmt: The risk with this pattern is that it invites more flexibility in the mixin than in required to get this feature functional. Instead of allowing the mixin arbitrary flexibility in `_pancakeInfinitySettleDelta` to call back into the base's `swapToClManager` and `swapToBinManager` (inversion of control) having a `virtual pure` function that simply returns the address of the CL manager or Bin manager (or revert) removes the footgun and (potentially, should be measured) improves `solc`'s ability to optimize.

I would expect that that change would also improve the amount of boilerplate required, potentially allowing us to move the vault/manager address selection into the chain-specific mixin. Of course, that totally breaks everything if there are 2 PancakeInfinity forks on the same chain

src/core/pancakeInfinityForks/OrvexCL.sol
```diff
@@ -0,0 +1,30 @@
...
+import {PancakeInfinityBase, PoolKey} from "../PancakeInfinity.sol";
+import {BalanceDelta} from "../UniswapV4Types.sol";
+import {revertUnknownPoolManagerId} from "../SettlerErrors.sol";
+
+address constant orvexVault = 0xFe7E25dE55e5cBbEcCcb661F3679F873f72B9b0D;
+address constant orvexClManager = 0xd01C774d4A66408326Bc65728Ac5Ae5aAf004032;
+
+abstract contract OrvexCL is PancakeInfinityBase {
+    function _PANCAKE_INFINITY_VAULT() internal pure override returns (address) {
+        return orvexVault;
```
duncancmt: You should check if `solc` is smart enough to optimize this constant or whether it needs this to be a literal in order to optimize.

src/core/PancakeInfinity.sol
```diff
@@ -593,4 +573,48 @@ abstract contract PancakeInfinity is SettlerSwapAbstract {
...
     }
+
+    // Each fork mixin in `pancakeInfinityForks/` defines its binding of pool manager IDs to pool
+    // managers by overriding this function with one that routes each recognized ID to
+    // `swapToClManager` or `swapToBinManager` with the appropriate pool manager's address, and
+    // calls `revertUnknownPoolManagerId` otherwise. A fork that lacks one of the pool manager
+    // types simply omits that route, which also omits the corresponding swap code from the
+    // compiled contract.
+    function _dispatchPancakeInfinity(
+        uint8 poolManagerId,
```
duncancmt: When possible, types should be widened to a full word to avoid `solc`'s wasteful cleaning of short types

duncancmt (approved):
This looks good, but the flexibility of the pattern (inversion of control) makes me worry about whether this introduces the possibility of future footguns. I think a defensively-programmed variation may also improve gas efficiency, but this is just a hunch.

src/core/pancakeInfinityForks/OrvexCL.sol
e1Ru1o: both approaches give exact same sizes

duncancmt (approved):

PR #626, opened 2026-08-18
Add scripts for upgrade safe timelock operations

duncancmt:


sh/submit_deployer_upgrade.sh
```diff
@@ -121,21 +121,19 @@ cd "$project_root"
 
 . "$project_root"/sh/common.sh
 
-echo 'Duncan wrote this for his own use' >&2
-echo 'If you are not Duncan, you are going to have a bad time' >&2
-echo 'If you are not using a frame wallet, doubly so' >&2
+die 'Duncan wrote this for his own use' \
+    'If you are not Duncan, you are going to have a bad time' \
+    'If you are not using a frame wallet, doubly so'
```
e1Ru1o, line 126: This one should stay as `echo` or the script should be removed

sh/common_safe_deployer.sh
```diff
@@ -1,3 +1,954 @@
...
+
+    cast concat-hex "$_pack_safe_confirmations_static" "$_pack_safe_confirmations_dynamic"
+}
+
+function _safe_confirmation_caller {
+    if [[ ${SAFE_GUARD_OVERRIDE:-${safe_guard:-null}} != [nN][uU][lL][lL] ]] ; then
+        cast to-checksum "$safe_guard"
+    elif [[ ${safe_signature_executor:-null} != [nN][uU][lL][lL] ]] ; then
+        cast to-checksum "$safe_signature_executor"
+    elif [[ ${signer:-null} = [nN][uU][lL][lL] ]] ; then
```
e1Ru1o: 🤔 
```suggestion
    elif [[ ${signer:-null} != [nN][uU][lL][lL] ]] ; then
```

sh/common_safe.sh
```diff
@@ -33,6 +40,43 @@ declare -a owners_array
...
+declare -r upgrade_safe_address
+
+declare safe_guard
+if [[ $installed_safe_guard = "$(cast address-zero)" ]] ; then
+    if [[ ${configured_safe_guard:-null} != [nN][uU][lL][lL] ]] ; then
+        die 'Safe '"$safe_address"' has no Guard installed, but governance.timelock says it has '"$configured_safe_guard"' for '"$chain_name"
+    fi
+elif [[ $(cast to-checksum "$safe_address") != "${upgrade_safe_address:-null}" ]] ; then
+    die 'Safe '"$safe_address"' is not the upgrade Safe, but has an installed Guard '"$installed_safe_guard"
+elif [[ $configured_safe_guard = "$(cast address-zero)" ]] ; then
```
e1Ru1o: Seems like this should be
```suggestion
elif [[ ${configured_safe_guard:-null} = [nN][uU][lL][lL] ]] ; then
```

sh/common_safe_deployer.sh
e1Ru1o: 🤔 
```suggestion
    elif [[ ${signer:-null} != [nN][uU][lL][lL] ]] ; then
```

test/0.8.25/SafeGuard.t.sol
```diff
@@ -1068,3 +1076,300 @@ contract TestSafeGuard is Test {
...
+        uint256 safeTxGas;
+        uint256 baseGas;
+        uint256 gasPrice;
+        address gasToken;
+        address payable refundReceiver;
+        uint256 nonce;
+    }
+
+    function setUp() public {
+        vm.createSelectFork(vm.envString("MAINNET_RPC_URL"), 23183520);
```
e1Ru1o, line 1113:
```suggestion
        vm.createSelectFork("mainnet", 23183520);
```

e1Ru1o (approved):

sh/submit_deployer_upgrade.sh
duncancmt: I'm keeping it in case we ever have to do another `Deployer` upgrade, but that's something that should happen so rarely that I would like to keep this around to prevent fatfingering. It's fine that the operator has to remove this guard before the script can run. Doing a `Deployer` upgrade is already an extremely manual process anyways

PR #628, opened 2026-08-21
Fix Immunefi bug reports 88903 and 89191 involving `CrossChainReceiverFactory` and partial proportional sells.

duncancmt:
There is no production impact from these two bug reports and their fixes. We are fixing this out of an abundance of caution.


e1Ru1o (approved):

src/core/UniswapV3Fork.sol
jparklev, line 350: tiny nit: this references `_updateSwapCallbackData`, which was removed

src/core/BalancerV3.sol
```diff
@@ -405,11 +405,11 @@ abstract contract BalancerV3 is SettlerSwapAbstract, FreeMemory {
...
 
         (uint256 amountIn, uint256 amountOut) = IBalancerV3Vault(msg.sender).unsafeSwap(swapParams);
-        unchecked {
-            // `amountIn` is always exactly `swapParams.amountGiven`
-            NotePtr sell = state.sell();
-            sell.setAmount(sell.amount() - amountIn);
-        }
+        // `amountIn` is always exactly `swapParams.amountGiven`, but `swapParams.amountGiven` can
+        // exceed `sell.amount()` if `bps` exceeds `BASIS`
+        NotePtr sell = state.sell();
```
jparklev, line 410: Does `_erc4626WrapUnwrap` suffer from this same issue? `bps` can be > `BASIS` and there's an unchecked sub

PR #631, opened 2026-08-25
Add `BEBOP` on RobinHood chain

duncancmt:


test/integration/RobinHood.t.sol
```diff
@@ -0,0 +1,58 @@
...
+        internal
+        view
+        override(BasePairTest, PancakeInfinityTest)
+        returns (uint160)
+    {
+        return super.sqrtPriceLimitX96(sellToken, buyToken);
+    }
+
+    function _testName() internal pure override returns (string memory) {
+        return "Orvex-WETH-NVDA";
```
e1Ru1o, line 40: All looks good but this that is generating `bebop_Orvex` snapshots. When it was a single contract it made sense but right right thing should be to overrided the pancake infinity part of the snapshot and replace it with Orvex

e1Ru1o (approved):

test/integration/RobinHood.t.sol
e1Ru1o: Maybe this #632

PR #633, opened 2026-08-25
Fix stack too deep in Plasma and Sonic

e1Ru1o:


src/utils/UnsafeMath.sol
```diff
@@ -155,4 +155,13 @@ library Math {
             r := sub(xor(sub(x, y), m), m)
         }
     }
+
+    function sub(uint256 x, uint256 y) internal pure returns (uint256 r) {
+        assembly ("memory-safe") {
+            r := sub(x, y)
+        }
```
duncancmt, line 162: Does this have to be assembly or can it be a normal `unchecked` block?

Also, nit, maybe this function should be named `checkedSub` for avoidance of ambiguity?

duncancmt (approved):

duncancmt (approved):

PR #634, opened 2026-08-25
Add FluxPool action on BNB

jparklev:


src/core/FluxPool.sol
```diff
@@ -0,0 +1,83 @@
...
+        unchecked {
+            sellAmount = sellToken.fastBalanceOf(address(this)) * ppm / BASIS;
+        }
+
+        uint256 balanceBefore = buyToken.fastBalanceOf(address(this));
+        _setOperatorAndCall(
+            FLUX_SWAP,
+            abi.encodeCall(
+                IFluxSwap.swapWithCallback,
+                (poolId, zeroForOne, sellAmount, minBuyAmount, address(this), abi.encodePacked(sellToken, sellAmount))
```
duncancmt: Can this action grow a `recipient` parameter to enable direct-to-user custody optimization? Does this action return a transferred amount? If not, then the `minBuyAmount` parameter should be removed

src/core/FluxPool.sol
```diff
@@ -0,0 +1,83 @@
...
+interface IFluxSwap {
+    function swapWithCallback(
+        bytes32 poolId,
+        bool zeroForOne,
+        uint256 amountIn,
+        uint256 amountOutMinimum,
+        address swapReceiver,
+        bytes calldata callbackData
+    ) external returns (uint256 amountOut);
+}
```
duncancmt, line 20: Not sure if this is needed, but you can probably save some bytes/gas by doing the encoding for this call in assembly.

src/core/FluxPool.sol
```diff
@@ -0,0 +1,83 @@
...
+            ),
+            uint32(IFluxSwapCallback.fluxSwapCallback.selector),
+            _fluxSwapCallback
+        );
+        uint256 buyAmount = buyToken.fastBalanceOf(address(this)) - balanceBefore;
+        if (buyAmount < minBuyAmount) revertTooMuchSlippage(buyToken, minBuyAmount, buyAmount);
+    }
+
+    function _fluxSwapCallback(bytes calldata data) private returns (bytes memory) {
+        if (data.length != 0xc0) revertConfusedDeputy();
```
duncancmt: Is this check required? Under what scenario might it fire?

src/core/FluxPool.sol
```diff
@@ -0,0 +1,83 @@
...
+        bool zeroForOne,
+        IERC20 buyToken,
+        uint256 minBuyAmount
+    ) internal {
+        uint256 sellAmount;
+        unchecked {
+            sellAmount = sellToken.fastBalanceOf(address(this)) * ppm / BASIS;
+        }
+
+        uint256 balanceBefore = buyToken.fastBalanceOf(address(this));
```
duncancmt: This is the forbidden slippage checking pattern. Please delete.

src/core/FluxPool.sol
```diff
@@ -0,0 +1,83 @@
...
+        // Equivalent Solidity pseudocode:
+        // (tokenToPay, amountToPay, bytes memory callbackData) = abi.decode(data, (IERC20, uint256, bytes));
+        // sellToken = IERC20(address(bytes20(callbackData[:20]))); sellAmount = uint256(bytes32(callbackData[20:]));
+        assembly ("memory-safe") {
+            tokenToPay := calldataload(data.offset)
+            amountToPay := calldataload(add(0x20, data.offset))
+            sellToken := shr(0x60, calldataload(add(0x80, data.offset)))
+            sellAmount := calldataload(add(0x94, data.offset))
+        }
+        if (tokenToPay != sellToken || amountToPay != sellAmount) revertConfusedDeputy();
```
duncancmt: If we can't determine that this check is unnecessary, then we probably shouldn't do this integration.

The `sellAmount` check is definitely unnecessary. Either the `amountToPay` is correct (inclusive of partial fills) or it is untrustworthy and this implementation is suspect.

If the check _is_ necessary, and for some reason we should still proceed with the integration, then this should use `FastLogic`

src/core/FluxPool.sol
```diff
@@ -0,0 +1,83 @@
...
+
+        sellToken.safeTransfer(FLUX_VAULT, sellAmount);
```
duncancmt: Do we need to hardcode `FLUX_VAULT`? Or do we get it as an argument somewhere? If we're supposed to retrieve it via an accessor on `FLUX_SWAP`, then it's fine to hardcode.

test/integration/FluxPool.t.sol
```diff
@@ -0,0 +1,100 @@
...
+
+        _setAllowed(FLUX_QUOTE_CURVE, QUOTE_CURVE_ALLOWLIST_SLOT, true);
+        _setAllowed(FLUX_VAULT, VAULT_ALLOWLIST_SLOT, true);
+    }
+
+    function testFluxPool() public {
+        uint256 vaultBalanceBefore = USDT.balanceOf(FLUX_VAULT);
+        snapStartName("settler_fluxPool");
+        _executeFluxPool();
+        snapEnd();
```
duncancmt, line 69: Gas profiling span too wide

duncancmt (commented):
@jparklev I assume that you have checked that the privately provided source code recompiles to exactly the deployed bytecode, and that the deployed source passes the callback unchanged? I haven't examined it in anything resembling the level of scrutiny that you have.

jparklev:
> I assume that you have checked that the privately provided source code recompiles to exactly the deployed bytecode, and that the deployed source passes the callback unchanged?

Yes, correct. They have admin control elsewhere, but callback comes back unchanged. Addressing this review is still WIP, and i'll share the contract verification tomorrow

src/core/FluxPool.sol
duncancmt: I saw that you used `FastLogic` here. That looks good. But the other part of the question remains unaddressed: what about the implementation of the pool/prop itself makes this check necessary?

src/core/FluxPool.sol
```diff
@@ -0,0 +1,76 @@
...
+        // Equivalent Solidity pseudocode:
+        // (tokenToPay, amountToPay, bytes memory callbackData) = abi.decode(data, (IERC20, uint256, bytes));
+        // sellToken = IERC20(address(bytes20(callbackData[:20]))); sellAmount = uint256(bytes32(callbackData[20:]));
+        assembly ("memory-safe") {
+            tokenToPay := calldataload(data.offset)
+            amountToPay := calldataload(add(0x20, data.offset))
+            sellToken := shr(0x60, calldataload(add(0x80, data.offset)))
+            sellAmount := calldataload(add(0x94, data.offset))
+        }
+        if ((tokenToPay != sellToken).or(amountToPay != sellAmount)) revertConfusedDeputy();
```
duncancmt: Can you add a comment explaining why this check is necessary?

src/core/FluxPool.sol
```diff
@@ -0,0 +1,76 @@
...
+            amountToPay := calldataload(add(0x20, data.offset))
+            sellToken := shr(0x60, calldataload(add(0x80, data.offset)))
+            sellAmount := calldataload(add(0x94, data.offset))
+        }
+        // The quote curve behind FLUX_SWAP is unverified and owner-swappable, so we never pay the
+        // amount it demands. Today it echoes `sellAmount`; a future curve that partial-fills would
+        // make us overpay silently, because the vault only checks `balance >= expected`.
+        if (amountToPay != sellAmount) revertConfusedDeputy();
+
+        sellToken.safeTransfer(FLUX_VAULT, sellAmount);
```
duncancmt:
```suggestion
        if (amountToPay > sellAmount) revertConfusedDeputy();

        sellToken.safeTransfer(FLUX_VAULT, amountToPay);
```

`revertConfusedDeputy` should probably be a different, more descriptive revert reason

src/core/FluxPool.sol
```diff
@@ -0,0 +1,76 @@
...
+        );
+    }
+
+    function _fluxSwapCallback(bytes calldata data) private returns (bytes memory) {
+        uint256 amountToPay;
+        IERC20 sellToken;
+        uint256 sellAmount;
+        // Decode the callback and packed token/amount without allocating memory.
+        // Equivalent Solidity pseudocode:
+        // (, amountToPay, bytes memory callbackData) = abi.decode(data, (IERC20, uint256, bytes));
```
duncancmt, line 71: Is `tokenToPay` controlled by the curve or `FLUX_SWAP`? If it's the latter, can't we read it out of the callback instead of having to nest it inside the `callbackData`?

src/core/FluxPool.sol
duncancmt: bump. can we add `recipient` to enable direct-to-taker?

duncancmt (commented):
nice!

jparklev:
Will respond to the remaining comments here tomorrow

src/core/FluxPool.sol
jparklev: 👌 went with a new `ExcessiveSellAmount(sellToken, maxSellAmount, sellAmount)`

src/core/FluxPool.sol
jparklev: No, I don't think so :/ At least in their v1. FluxSwap requires `swapReceiver == msg.sender` and the vault sends output token to that address

src/core/FluxPool.sol
jparklev: Good shout -- reduced bytes down ~ 70B and saved a bit on gas as well. taken here: https://github.com/0xProject/0x-settler/pull/634/changes/0c7f56d9599f1698395d95aabe652f3ee037c0c4

src/core/FluxPool.sol
jparklev: The curve, unfortunately. FluxSwap just relays `settlement.tokenIn` from `quoteAndConsume`

src/core/FluxPool.sol
jparklev: no accessor , `vault` is `private` on FluxSwap and the owner can `setVault`. we can make this an arg rather than hardcode if you'd prefer

src/core/FluxPool.sol
duncancmt: OK, cool. Probably worth a comment so that this doesn't come up again

src/core/FluxPool.sol
```diff
@@ -0,0 +1,83 @@
...
+            sellAmount = sellToken.fastBalanceOf(address(this)) * ppm / Constants.BASIS;
+        }
+
+        // Equivalent to `abi.encodeCall(IFluxSwap.swapWithCallback, (poolId, zeroForOne, sellAmount, minBuyAmount,
+        // address(this), abi.encodePacked(sellToken, sellAmount)))`, but tightly packed and canonicalizing `zeroForOne`.
+        bytes memory data;
+        assembly ("memory-safe") {
+            data := mload(0x40)
+            mstore(add(0x104, data), shl(0x60, sellToken))
+            mstore(add(0x118, data), sellAmount)
```
duncancmt: You can adjust the offset used for `sellToken` and then use the `0x34` write to clobber any dirty bits, eliminating the `shl`

duncancmt (approved):

src/core/FluxPool.sol
duncancmt: no, this is fine

jparklev:
Thank you for the reviews 🫡

duncancmt (approved):

jparklev:
Update: The Flux team has said that they are pushing a new Quoter contract ASAP, to deal with some things they are seeing on-chain, and also updating their sdk. We don't use the Quoter in the action, and they've said `FluxSwap` itself is not changing (the only change that will need to be made is in the fork tests). Even so, holding off on merging until we see those updates from them

duncancmt (approved):

PR #639, opened 2026-08-27
Only check for Safe{Wallet} Guard presence when upgrade Safe is configured

duncancmt:


sh/common_safe.sh
```diff
@@ -40,42 +40,43 @@ declare -a owners_array
...
-else
-    safe_guard="$configured_safe_guard"
+    declare -r configured_safe_guard
+
+    declare safe_guard
+    if [[ $installed_safe_guard = "$(cast address-zero)" ]] ; then
+        if [[ ${configured_safe_guard:-null} != [nN][uU][lL][lL] ]] ; then
+            die 'Safe '"$safe_address"' has no Guard installed, but governance.timelock says it has '"$configured_safe_guard"' for '"$chain_name"
+        fi
+    elif [[ $(cast to-checksum "$safe_address") != "${upgrade_safe_address:-null}" ]] ; then
```
e1Ru1o: This is inside an `if` that checks that they are equal

sh/common_safe.sh
```diff
@@ -40,42 +40,43 @@ declare -a owners_array
...
 fi
 declare -r upgrade_safe_address
 
-declare safe_guard
-if [[ $installed_safe_guard = "$(cast address-zero)" ]] ; then
+if [[ $(cast to-checksum "$safe_address") = "$upgrade_safe_address" ]] ; then
+    declare installed_safe_guard
+    installed_safe_guard="$(cast call --rpc-url "$rpc_url" "$safe_address" 'getStorageAt(uint256,uint256)(bytes)' "$(cast keccak 'guard_manager.guard.address')" 1)"
+    installed_safe_guard="$(cast parse-bytes32-address "$installed_safe_guard")"
+    declare -r installed_safe_guard
```
e1Ru1o: you might want to have this one outside of the `if` to check that is `0x0` in the case `safe != upgradeSafe`

e1Ru1o (changes requested):
I think `common_safe.sh` still need changes.

e1Ru1o:
And seems like there are still old uses of `BASIS`, without the `Constants` import which are breaking the test

e1Ru1o (approved):

PR #640, opened 2026-08-28
feat: integrate Deepstate settlement on Robinhood Chain

dangerousfood:
## Summary

Adds a chain-scoped `DEEPSTATE` self-funded action to Robinhood Settler for the canonical Deepstate V1 engine at `0x6cf19308C22FC82ea620Fa0B3E94948d20f27B96`.

- executes one Deepstate book fill per action
- derives the maximum input from the current Settler balance using standard ppm semantics
- derives ask/bid direction and sorted `token0`/`token1` from the action assets
- sizes asks directly from the selected token0 balance
- sizes bids from the selected token1 budget using a downward-rounded Q128 reciprocal of the encoded limit price
- forces `noRest = true` and `fillOrKill = false`
- leaves unmatched input in Settler for subsequent route actions
- supports native and ERC-20 input/output
- remains scoped to Robinhood Settlers

## Router encoding

For a bid, `inversePriceX128` is `floor(2**128 / price(tick))`, where `price(tick) = 2 ** (96 * tick / 2**31)`. It is ignored for asks. The value is derived from the selected limit tick during route construction; it does not require an onchain quote.

Each action contains one pair/book. A route can compose multiple independently resized actions without forwarding a caller-authored Deepstate fill array.

## Security boundaries

- requires the pinned engine to contain code when production Settlers are deployed
- restricts the engine from generic `BASIC` calls so validation and `noRest` cannot be bypassed
- grants only the selected ppm budget as a temporary ERC-20 allowance and clears it after settlement
- verifies Settler and engine input balance deltas agree exactly
- rejects fee-on-transfer, rebasing, or otherwise inexact ERC-20 input behavior
- rejects zero-address ERC-20 encodings and identical assets
- forwards only the selected native-input budget and no unrelated native balance
- saturates packed quantities at `uint160` without corrupting the signed tick field
- relies on the transaction-level Settler slippage check for final output protection

## Verification

- Deepstate unit suite: 19/19 passed
- ask resizing fuzzing: 100,000 runs passed
- bid resizing and 512-bit multiplication fuzzing: 100,000 runs passed
- live Robinhood fork at block `52,936,000`: 2/2 passed against the deployed engine
  - balance-sized ask with unmatched input
  - ppm-sized bid at an exact nonzero price against multiple resting asks
  - protocol-fee, allowance-clearing, and `noRest` assertions
- broader unit/FV/invariant run: 213 tests passed
- SafeGuard suite under the required London profile: 20/20 passed
- Uniswap V4 invariant: 20,000 calls with zero reverts
- pinned 0.8.25, 0.8.28, and 0.8.34 builds: passed
- scoped format/lint and action-selector checks: passed
- gas snapshot comparison: no regression

## Gas Optimization

- Deepstate execution gas snapshot: 170,523 gas (16.84% below the prior implementation)
- `RobinHoodSettler`: 21,002 runtime bytes, 3,574-byte EIP-170 margin
- `RobinHoodSettlerIntent`: 19,696 runtime bytes, 4,880-byte margin
- `RobinHoodSettlerMetaTxn`: 18,889 runtime bytes, 5,687-byte margin
- the action constructs one fixed-size `fill` call and avoids copying a dynamic fill array


dangerousfood:
Addressed the fill-resizing concern in `f3792193`.

The action now represents exactly one Deepstate book fill and sizes it from the current Settler balance at execution time. Asks use the selected token0 amount directly. Bids convert the selected token1 budget with `floor(2**128 / price(tick))`; that reciprocal comes from the already-selected limit tick, so no onchain quote is needed.

Every call hardcodes `noRest = true`. Unmatched input stays in Settler for later route actions. For ERC-20 input, the engine receives an exact temporary allowance equal to the ppm budget, which is cleared afterward and independently prevents overspending even if malformed calldata supplies an inconsistent reciprocal.

I added 100,000-run ask and bid fuzz tests, overflow/packing and exact-spend tests, plus live deployed-engine coverage for both an ask and a nonzero-price bid consuming multiple resting asks. The full local suite is green, and the Robinhood Settler retains 3,574 bytes of EIP-170 margin.

PR #641, opened 2026-09-01
Add `SELECT` action

duncancmt:
Copied from 0xProject/0x-settler-private#6 . Most of this work is from @jparklev 

- add the SELECT action
- no VIP action support within SELECT
- no on-chain _measured_ Best-of-n (data didn't support meaningful uplift from production-representative quote -> execute delays). however, the interface we have here can also support best-of-n, with a relatively simple change to the internals of `_select` if we choose to add it back (same selector and encoding)


src/ISettlerActions.sol
```diff
@@ -224,6 +224,18 @@ interface ISettlerActions {
     // Post-req: Payout
     function BASIC(address sellToken, uint256 ppm, address pool, uint256 offset, bytes calldata data) external;
 
+    /// @dev Tries funded action sequences (`candidates`) in order, and commits the first to meet
+    ///      its score `target`. A score is the increase in the `token` balance held by Settler. A
+    ///      zero `target` commits any non-reverting candidate. Native asset output only scores if
+    ///      wrapped to wrapped-native in the candidate actions. Candidates must not contain
+    ///      `CHECK_SLIPPAGE`. `targets` are the per-candidate minimum outputs that terminate the
```
jparklev, line 231: For reviewers: we’ll add `NATIVE_CHECK` to this comment for clarity. Both `CHECK_SLIPPAGE` and `NATIVE_CHECK` should stay outside SELECT candidates

PR #642, opened 2026-09-02
Deepstate integration on Robinhood Chain

e1Ru1o:
## Summary

Integrates the Deepstate on-chain limit-order engine on Robinhood Chain as the `DEEPSTATE` action.

- One book fill per action, funded by the Settler's current `sellToken` balance via `ppm`. Multi-hop routes are chained actions, each sized from what actually arrived.
- `token0`/`token1` ordering and the bid/ask side are derived on-chain from the two tokens, with the ETH sentinel mapped to Deepstate's `address(0)`.
- Asks are sized directly by the sell amount. Bids are sized through a caller-supplied `inversePriceX128` (Q128 `token0` per `token1`), since the engine only accepts `token0` quantities. An adverse price move reverts on the engine's `transferFrom`.
- `noRest` is always forced on so unmatched quantity stays in the Settler for later actions instead of resting as an order the Settler could never cancel. `fillOrKill` is off; the final `AllowedSlippage` check protects the taker.
- The call is hand-encoded to `fill`. No refund logic, no balance reconciliation, no restricted-target override, no custom errors.

## Gas Optimization

- Replaces the previous `fillRoute` path (route array allocation, three balance reads, temporary approve/reset, refund transfer) with a single hand-encoded `fill` call, one balance read, and a standing max approval via `safeApproveIfBelow`.
- `RobinHoodSettler` runtime size drops from 22,928 to 20,427 bytes versus the previous Deepstate commit on this branch.
- No gas snapshot for the action yet. The fork test writes `settler_deepstate_robinhood` when run in CI; the stale snapshot from the previous implementation was removed.

## Testing

- `test/integration/RobinHood.t.sol` runs the production `RobinHoodSettler` against the live engine on a Robinhood Chain fork. Not run locally (no `ROBINHOOD_MAINNET_RPC_URL`), so CI is the first execution. It currently uses tick 0 and should gain a negative-tick and a bid case before merge.
- The rest of the unit suite passes locally.

## Open items

- The live engine at `0x6cf19308C22FC82ea620Fa0B3E94948d20f27B96` has not been bytecode-compared against the upstream repository.
- A quantity at or above 2^160 would overflow into the tick field of the packed order. Left as GIGO, matching Hanji's truncating cast; say if a revert is preferred.


src/ISettlerActions.sol
```diff
@@ -7,7 +7,6 @@ interface ISettlerActions {
     /// VIP actions should always start with `recipient` address and the `permit` from the taker
     /// followed by all the other parameters to ensure compatibility with `executeWithPermit` entrypoint.
     /// `minBuyAmount`/`amountOutMin` should always be the last parameter.
-
```
duncancmt, line 10: plz fix

src/ISettlerActions.sol
```diff
@@ -224,6 +223,23 @@ interface ISettlerActions {
     // Post-req: Payout
     function BASIC(address sellToken, uint256 ppm, address pool, uint256 offset, bytes calldata data) external;
 
+    /// @dev Fills one Deepstate book with the current `sellToken` balance. `token0` is the lower of the two
+    /// token addresses with native ETH as `address(0)`; selling `token0` is an ask and selling `token1` is a
```
duncancmt: Please use ERC-7528

src/ISettlerActions.sol
```diff
@@ -224,6 +223,23 @@ interface ISettlerActions {
...
+    /// bid. Unmatched quantity is discarded rather than rested, leaving it as `sellToken` for later actions.
+    /// @param epoch Book epoch to match against. The book must already be initialized.
+    /// @param tick Signed logarithmic limit price of `token1` per `token0`, `2 ** (96 * tick / 2**31)`.
+    /// @param inversePriceX128 Q128 quantity of `token0` bought per unit of `token1` sold. Sizes a bid; ignored for an ask.
+    // Pre-req: Funded
+    // Post-req: Payout
+    function DEEPSTATE(
+        address sellToken,
+        uint256 ppm,
```
duncancmt, line 242: The comment above implies that this argument is unused. Which is it?

src/ISettlerActions.sol
```diff
@@ -224,6 +223,23 @@ interface ISettlerActions {
     // Post-req: Payout
     function BASIC(address sellToken, uint256 ppm, address pool, uint256 offset, bytes calldata data) external;
 
+    /// @dev Fills one Deepstate book with the current `sellToken` balance. `token0` is the lower of the two
+    /// token addresses with native ETH as `address(0)`; selling `token0` is an ask and selling `token1` is a
+    /// bid. Unmatched quantity is discarded rather than rested, leaving it as `sellToken` for later actions.
+    /// @param epoch Book epoch to match against. The book must already be initialized.
+    /// @param tick Signed logarithmic limit price of `token1` per `token0`, `2 ** (96 * tick / 2**31)`.
+    /// @param inversePriceX128 Q128 quantity of `token0` bought per unit of `token1` sold. Sizes a bid; ignored for an ask.
```
duncancmt: I think this parameter needs a better explanation. Since we have both `tick` and `inversePriceX128`, an explanation of how they're different and when to use which one is important

src/core/Deepstate.sol
```diff
@@ -0,0 +1,119 @@
...
+                sellAmount = (address(this).balance * ppm).unsafeDiv(Constants.BASIS);
+            } else {
+                sellAmount = (sellToken.fastBalanceOf(address(this)) * ppm).unsafeDiv(Constants.BASIS);
+                sellToken.safeApproveIfBelow(address(ROBINHOOD_DEEPSTATE), sellAmount);
+            }
+        }
+
+        // Deepstate designates native ETH as `address(0)`, so it always sorts as `token0`.
+        address sellAsset = sendNative.ternary(address(0), address(sellToken));
+        address buyAsset = (address(buyToken) == Constants.ETH_ADDRESS).ternary(address(0), address(buyToken));
```
duncancmt: `ternary` with zero can probably be golfed. See if that's important. Also I note the inconsistent use of ERC-7528 here. That should be homogenized.

src/core/Deepstate.sol
```diff
@@ -0,0 +1,119 @@
...
+        bool isBid = sellAsset > buyAsset;
+        (address token0, address token1) = isBid.maybeSwap(sellAsset, buyAsset);
+
+        // Orders are sized in `token0`. A bid spends `token1`, so the sell amount is converted through the
+        // caller's price. Rounding down keeps the engine's debit within the sell amount; the engine's own
+        // per-order rounding is absorbed by the caller's choice of `tick`.
+        uint256 quantity = sellAmount;
+        if (isBid) {
+            (uint256 hi, uint256 lo) = tmp().omul(sellAmount, inversePriceX128).into();
```
duncancmt: Wait.... is `inversePriceX128` the off-chain computed realized price? That's not a good design because then the filling amount depends on slippage.

src/core/Deepstate.sol
```diff
@@ -0,0 +1,119 @@
...
+            quantity = (hi << 128) | (lo >> 128);
```
duncancmt: `512Math` has shifting functions available. Use them.

src/core/Deepstate.sol
```diff
@@ -0,0 +1,119 @@
...
+    ) internal {
+        // Manually ABI-encode the call to avoid allocating and copying a `FillParams` struct. Equivalent to:
+        //     deepstate.fill{value: value}(
+        //         IDeepstateV1.FillParams(token0, token1, epoch, bytes32(uint256(uint32(tick)) << 224 | quantity << 64), isBid, true, false)
+        //     );
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            mstore(ptr, 0x5d6222ab) // selector for `fill((address,address,uint256,bytes32,bool,bool,bool))`
+            mstore(add(0x20, ptr), and(0xffffffffffffffffffffffffffffffffffffffff, token0))
+            mstore(add(0x40, ptr), and(0xffffffffffffffffffffffffffffffffffffffff, token1))
```
duncancmt: Avoid masking by more careful construction of the ABIencoding. The selector should contain the padding for `token0` and then `token0` should be `shl`'d to provide the padding for `token1`. See `SafeTransferLib` for examples.

src/ISettlerActions.sol
e1Ru1o: I hate AI always messing that line up

src/ISettlerActions.sol
e1Ru1o: It is used besides this comment maybe not describing that

src/ISettlerActions.sol
e1Ru1o: Check if it looks better now

src/core/Deepstate.sol
e1Ru1o: This value, is directly related to the `tick`, I don't see a way to go around it. It is provided to avoid doing the tick match in the contract

src/ISettlerActions.sol
```diff
@@ -7,7 +7,7 @@ interface ISettlerActions {
     /// VIP actions should always start with `recipient` address and the `permit` from the taker
     /// followed by all the other parameters to ensure compatibility with `executeWithPermit` entrypoint.
     /// `minBuyAmount`/`amountOutMin` should always be the last parameter.
-
+
```
duncancmt, line 10: plz fix.

src/ISettlerActions.sol
```diff
@@ -224,6 +224,28 @@ interface ISettlerActions {
...
+    /// both `sellToken` and `buyToken`. Unmatched quantity is discarded rather than rested, so it stays in the
+    /// Settler as `sellToken` for later actions.
+    /// @param epoch Book epoch to match against. The book must already be initialized.
+    /// @param tick Limit price enforced by the engine: `2 ** (96 * tick / 2**31)` units of the higher-addressed
+    /// token per unit of the lower-addressed token.
+    /// @param inversePriceX128 `floor(2**128 / price(tick + 1))`, the Q128 reciprocal of the limit price one tick
+    /// past `tick`. Only used when selling the higher-addressed token (a bid): Deepstate sizes bids in the
+    /// lower-addressed token, so the sell amount is converted through this value. Sizing at the limit price means
+    /// the engine can never take more than the sell amount; if the book is better than the limit, the difference
+    /// stays unspent. Ignored for an ask.
```
duncancmt, line 237: I still think this is a suspect design choice. If keeping this, I think this needs a better explanation of why it has to be "one tick past `tick`" because it's not obvious to me how that avoids introducing a fencepost error

src/core/Deepstate.sol
```diff
@@ -0,0 +1,125 @@
...
+        bool isBid = sellToken > buyToken;
+        (IERC20 token0, IERC20 token1) = isBid.maybeSwap(sellToken, buyToken);
+
+        // Orders are sized in `token0`. A bid spends `token1`, so the sell amount is converted through the
+        // reciprocal of the limit price, which bounds the engine's debit by the sell amount. The one wei the
+        // engine can add when it partially consumes an ask is absorbed by the caller taking `inversePriceX128`
+        // one tick past the limit.
+        if (isBid) {
+            (, sellAmount) = tmp().omul(sellAmount, inversePriceX128).ishr(128).into();
+        }
```
duncancmt, line 121: In contrast, #640 does:

```Solidity
sellToken.safeApprove(address(ROBINHOOD_DEEPSTATE), 0);
sellToken.safeApprove(address(ROBINHOOD_DEEPSTATE), maxSellAmount);
ROBINHOOD_DEEPSTATE.fastFill(0, token0, token1, epoch, tick, quantity, isBid);
sellToken.safeApprove(address(ROBINHOOD_DEEPSTATE), 0);
```

I haven't actually read the Deepstate code. Does Deepstate introspect the allowance amount and only pull up to the limit?

e1Ru1o: no, it doesn't introspect, it pull anything needed to pay the debt but, 640 was setting the `maxSellAmount` as a guard to make sure not more than that is pulled.

src/ISettlerActions.sol
e1Ru1o: I am going to do a final round on this and double check with Deepstate

src/core/Deepstate.sol
```diff
@@ -0,0 +1,128 @@
...
+            mstore(add(0xa0, ptr), lt(0x00, isBid))
+            mstore(add(0x80, ptr), or(shl(0xe0, tick), shl(0x40, quantity)))
+            mstore(add(0x60, ptr), epoch)
+            mstore(add(0x40, ptr), token1)
+            mstore(add(0x2c, ptr), shl(0x60, token0)) // Clears `token1`'s padding.
+            mstore(add(0x0c, ptr), 0x5d6222ab000000000000000000000000) // Selector for `fill((address,address,uint256,bytes32,bool,bool,bool))`, with `token0`'s padding.
+
+            if iszero(call(gas(), deepstate, value, add(0x1c, ptr), 0xe4, 0x00, 0x00)) {
+                returndatacopy(ptr, 0x00, returndatasize())
+                revert(ptr, returndatasize())
```
duncancmt:
```suggestion
                let ptr_ := mload(0x40)
                returndatacopy(ptr_, 0x00, returndatasize())
                revert(ptr_, returndatasize())
```
small golf

src/core/Deepstate.sol
```diff
@@ -0,0 +1,128 @@
...
+        uint256 epoch,
+        int32 tick,
+        uint256 inversePriceX128
+    ) internal {
+        bool sendNative;
+        assembly ("memory-safe") {
+            // DeepState uses address(0) for native token
+            let shiftedXor := shl(0x60, xor(sellToken, ETH_ADDRESS))
+            sellToken := mul(sellToken, lt(0x00, shiftedXor))
+            buyToken := mul(buyToken, lt(0x00, shl(0x60, xor(buyToken, ETH_ADDRESS))))
```
duncancmt:
```suggestion
            let shiftedXor := shl(0x60, xor(ETH_ADDRESS, sellToken))
            sellToken := mul(lt(0x00, shiftedXor), sellToken)
            buyToken := mul(lt(0x00, shl(0x60, xor(ETH_ADDRESS, buyToken))), buyToken)
```
put constants on the left

src/core/Deepstate.sol
duncancmt: this is gonna be a pain for the liquidity integrations guys, but OK

duncancmt (approved):

src/core/Deepstate.sol
duncancmt: can you document this _**thoroughly**_ in the changelog?

e1Ru1o: I get it, that it is going to be pain, but, either they put a price there or we walk the tree. Confirming with Deepstate that there is no other way to go about this

PR #650, opened 2026-09-11
fix: bound EulerSwap input limit by vault capacity

jparklev:
This is something we can fix. It's minor and in the worst case leads to a reverted trade. Settler's limit for an EulerSwap fill can come out a bit higher than what the pool takes.

But also, we're not using this action at all in the solver. Should we remove it instead?

### DIFF
--- a/src/core/EulerSwap.sol
+++ b/src/core/EulerSwap.sol
@@ -659,9 +659,13 @@ library EulerSwapLib {
             }
         }
 
+        // The pool deposits the fee too, so the headroom caps the gross input.
         unchecked {
-            inLimit = (inLimitFromOutLimit < inLimit).ternary(inLimitFromOutLimit, inLimit);
-            inLimit = (inLimit * 1e18).unsafeDiv(1e18 - p.fee());
+            uint256 netHeadroom = (inLimit * (1e18 - p.fee())).unsafeDiv(1e18);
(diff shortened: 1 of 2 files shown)


duncancmt (approved):
> But also, we're not using this action at all in the solver. Should we remove it instead?

I have no strong opinion either way. Do you think we'll ever _start_ using it? I believe we need to update the parameter layout for the version 2 that is currently live, if we do plan on using it

jparklev:
> > But also, we're not using this action at all in the solver. Should we remove it instead?
> 
> I have no strong opinion either way. Do you think we'll ever _start_ using it? I believe we need to update the parameter layout for the version 2 that is currently live, if we do plan on using it

Yea, after looking more into usage, i'd advocate for removing it entirely. I think support for even v2 is declining, and usage is low. I've messaged robert to get his sign off on that

Will close this PR in favor of a removal if so

Screenshot 2026-09-14 at 12 10 23 PM

duncancmt:
> Will close this PR in favor of a removal

if you do that, can you keep these fixes in the commit history so that we can dig them out if we need them in the future?

jparklev:
Closing in favor of https://github.com/0xProject/0x-settler/pull/655. See that PR description, and the comments above, for context. Fix commit cherry picked over

